PR #15, opened 2021-01-08
MOULa Textures

Hoikas:
Per discussion in OpenUru#dev discord.

Some things:
- I'm not sold on this being the best directory organization for the textures - input welcome
- There seem to be a few textures that have the string "Asset" prepended to them. Do we want to rename those to drop the spurious prefix?
- Several of these are from realMyst


DoobesURU:
My thoughts:

* I wonder if we couldn't divide things up by Age (aside from the textures with the "x" prefix...those could be part of a "common" directory)? It'd help Age builders like myself track down a particular texture I see in an Age.
* Since they're all technically "assets", I think we could probably drop that from the filename. 😉

dgelessus:
> * I wonder if we couldn't divide things up by Age (aside from the textures with the "x" prefix...those could be part of a "common" directory)?

If someone hasn't already done this yet, I would be happy to help out with that. Would it be of any use if I make a branch based on this PR, rename/sort the files there, and submit that somehow (as another PR, or just push it to my fork)?

Hoikas:
I'm not particularly convinced that sorting them by Age would wind up being a good idea, for a few reasons. Those mostly center around that although a texture may have a prefix indicating it's used in, for example, Ahnonay or the bug cave, you will still see it being used in other Ages. Further, if we ever get more Max files from Cyan, those Max files will be looking for these textures to be in a single directory. But that also argues against compressing them as PNGs and removing redundant prefixes, heh.

dgelessus:
> Further, if we ever get more Max files from Cyan, those Max files will be looking for these textures to be in a single directory.

Valid point - if the original Max files expect the textures with these exact names in a flat directory, then they should be preserved somewhere in that state. But for almost all other use cases, I think it would be better to have *some* sort of directory structure. Since this is Git, I think it's not a big issue to add extra commits that reorganize the files - if you need the original structure, you can just check out the original commit that added the textures (I suppose we could add a branch or tag to make it easier).

You're right that sorting by age isn't perfect, but it works well enough in most cases, and it's certainly better than a single flat directory. Also, we already *have* a grouping by age, from the file name prefixes, except that they are sometimes inconsistent/overlapping/missing. So I think it would be a natural improvement to group the textures into directories based on the ages/prefixes, but without the inconsistencies.

dgelessus:
Another thing I noticed - some of the TGAs that this PR adds are already in the repo under sources/KI and sources/avatar, along with max files that presumably use those textures. If we end up sorting the textures in this PR by age, it would probably make sense to reuse those directories, to avoid duplicating files, and to have all textures for each age/"topic" in one place.

Though if more original max files are added to this repo in the future, in the same way as the existing max files under sources (i. e. grouped thematically and with all textures directly next to the max files that use them), having some duplicated textures would be inevitable. For example, under sources/KI, there are already textures with nb01 and tldn prefixes. If the max files for Teledahn or the hoods were added in another directory, those textures would be duplicated.

Hoikas:
The good news regarding duplicated files is that they're all stored in the LFS by hash and a tiny pointer lives in-repo. I know that doesn't answer any questions, but I think it's a relevant point.

dgelessus:
Indeed, which means that duplicate files at least don't waste storage in the history and on the server. But when checked out, duplicate files (LFS or not) are actually duplicated on disk. A couple of duplicated textures won't be a big issue space-wise, but it means we can't for example have two alternative structures for the textures (one flat and one grouped) without doubling the size of the checkout.

Hoikas:
@dgelessus If you would like to start organizing the textures, feel free. You could open your own PR or I could push your branch to this PR - I'm not picky about how that happens. I was working on making a list of textures to *remove*. This is what I had:
```
00stone1
0Rm_panel2
0Rm_panel2side
0RmButton
al102
al103
al104
al105
al109
al110
AL201
AL202
AL203
AL204
AL205
AL206
AL207
AL208
AL209
AL210
AL211
am
anchor_off
anchor_on
arrow_off
ARROW_ON
arrow_moon
BB*
BIRD_OFF
BIRD_ON
book1
book2
catherine_paper
catherinepaper_sh
ClockFace
cross_off
CROSS_ON
DECK
DeleteBtn
drcHeekScores*
drcSafety01
drcSafety02
drcSafety03
drcWelcome
EYE_OFF
EYE_ON
gtghot200
hintpanel*
insect_off
insect_on
LEAF_OFF
LEAF_ON
lib*
linkbtn
lp*
marker*
may
MobileSign
MT_*
nb01ImagerPhoto*
number*
ot*
pa*
pb*
Pen
Pencil
pi1*
pi2*
Screenshot*
SNAKE_OFF
SNAKE_ON
SnowAgeLinkCover*
starmap*
StartPanel
tldnCrateSide01

warning: ShrmAgePict2 has color data on xparent pixels
```

dgelessus:
Alright, I've done some initial sorting on a branch in my fork: https://github.com/dgelessus/moul-assets/tree/textures

I sorted most things based on obvious file name prefixes and sorted most of the rest manually. Many of the textures under "Myst" have no prefix and I sometimes guessed only based on the naming pattern. For the textures where I really have no idea where they belong I made an "unknown" folder.

I also removed a few files that were byte-identical duplicates of others, e. g. most of the "Assets"-prefixed ones. I didn't delete any unique textures, though I moved some obviously not useful ones into separate folders.

Will probably still do some more sorting under the avatar and GUI textures, and perhaps under Myst to move away some of the junk there.

Any feedback/suggestions welcome 🙂

Hoikas:
A few initial observations:
- Try to avoid spaces in all path names (both files and directories). There are enough tools out there in the world that will error on these that we should probably not use them.
- Try to match exactly the age filename when applicable. In the case of multiple, use the most applicable So, for example:
    - `Garden`, not `Eder Kemo`
    - `Gira`, not `Eder Gira`
    - `Neighborhood`, not `Neighborhoods`
    - `Kadish`, not `Kadish Tolesa`
    - `city`, not `City`
    - etc
- Instead of `shared`, maybe `global`

dgelessus:
> * Try to avoid spaces in all path names (both files and directories). There are enough tools out there in the world that will error on these that we should probably not use them.

ugh. But yeah, you're right that it's probably safer to not use spaces. Any preference regarding dev_textures vs. dev-textures vs. devTextures?

> * Try to match exactly the age filename when applicable. In the case of multiple, use the most applicable

I started out doing that actually, but changed some of the names for clarity. "Garden" is a fun one especially - there are a number of textures with a "grdn" prefix, but most of them are actually related to the light garden in the hoods and not Eder Kemo! (But now that those are correctly sorted, that's less of a problem.)

> * `Neighborhood`, not `Neighborhoods`

That was intentional to make it clear that it also contains textures specific to Kirel (`Neighborhood02`) and not just generic hood textures. But if we're going for exact age names I can rename it back to `Neighborhood`.

> * `city`, not `City`

ho that's going to be a fun one to change with Windows case-insensitivity 😄

> * Instead of `shared`, maybe `global`

That's also intentional, to avoid confusion with textures that belong to GlobalAvatars, GlobalClothing, etc. The ones currently in `shared` are just used across multiple ages (e. g. journey cloths, maintainer mark) but not technically global.

Actually that's not true for the marker textures (which I currently put under `shared`). If we're going by data file names, I should make a separate GlobalMarkers folder for them.

dgelessus:
Relevant bits from the discussion on the OpenUru Discord about converting the TGA textures to PNG:

> **Adam {Hoikas}**: Need to be careful doing that [converting to PNG]. Several of the textures contain transparent pixels with color data -- many PNG compressors will throw this information away. Also, for what goes into the repo, we would ideally like to spend extra time on those images that do not need special consideration to run an extreme level of compression on them, for example, pngcrush.
> 
> [...]
> 
> **Adam {Hoikas}**: I'm not arguing against converting to PNG (I am in favor of it) - just saying that it needs to be done carefully with knowledge about what exactly the image is trying to be/do
> 
> [...]
> 
> **dgelessus**: I think it would make sense to keep just the original TGAs in the moul-assets repo and provide the PNG-converted versions as a separate download in a zip/7z archive. Then we have one version that's easily accessible for people who just want to look at the textures, and a full copy of the original unmodified assets for those who need them. There should be no need to keep the converted PNGs in version control, because they can always be regenerated from the original TGAs.

Hoikas:
Sorry for dropping out on this, @dgelessus. is your branch ready to be pushed to this PR?

dgelessus:
> is your branch ready to be pushed to this PR?

Good question, heh. I was planning to slightly reorganize the big folders where possible, e. g. separating the book/journal textures under GUI from the actual GUI textures, and perhaps figuring out *some* useful grouping for GlobalClothing. Then there's also the question if/how we should avoid duplication with the TGAs under the existing sources subdirectories. Haven't gotten around to doing any of this yet, because time.

IMO, none of this is critical though - the major rough sorting work is done, so feel free to use the current state of my branch in this PR if you want. Once I get around to doing the other changes I was thinking about, I can submit those as a separate PR.

Hoikas:
@Deledrius, @DoobesURU, @dpogue, Any OHBJECTIONS to this ordering?

DoobesURU:
> 
> 
> @Deledrius, @DoobesURU, @dpogue, Any OHBJECTIONS to this ordering?

Only the fact we're still putting an "h" in "objection". 😉 Other than that, this looks good!

dpogue:
:+1: from me. Some organization is better than no organization.

Hoikas:
Thanks for getting this done, @dgelessus!

PR #19, opened 2021-02-04
[Teledahn] Fix spurious workroom noise.

Hoikas:
Corrected the plSoundMsg in the workroom power stop responder to send the `kStop` command instead of `kPlay`, causing the screw turn noise to stop playing when the power is off. See original report on the [forum](https://forum.guildofwriters.org/viewtopic.php?p=75033#p75033).


Hoikas:
This change has been tested on Minkata and will be delivered to the MOULa pipeline Monday. Merging.

PR #20, opened 2021-02-11
Add Tre'bivdil to MOULa

DoobesURU:
Adds the Age of Tre'bivdil to MOULa.

Age Details:
* Name: Tre'bivdil
* Filename: trebivdil
* Creator: Lontahv
* Prefix: 10228

Files include 3 PRPs and one OGG.

Minor changes/additions from the original:
* Added ripples and wet footprints for the small pool.
* Adjusted sittingmod regions to prevent avatar "shakiness"

Some issues are present and are open to suggestions:
* There is an odd artifact in the waveset EnvMap if you look to the left of the Link in.
* Waveset does not allow visibility of the islands beneath the water line (possibly better this way?)
* The stairs down go a bit too far and show an unfinished section. Maybe block this?

Linking Book Def has already been added as of last major update and the Book in Chiso (`BookVis01` SDL value) should be ready to go.


Deledrius:
Is the waveset EnvMap bug a regression?

DoobesURU:
> Is the waveset EnvMap bug a regression?

Well, it seems fine in the Gehn version, but when exported with Korman, either with the original cubemap texture or DEM, it's present in both cases.

PR #21, opened 2021-02-20
Myst 1 to 1

DoobesURU:
A 1 to 1 conversion of the Myst Library in URU to Blender/Korman with new PRPs and blend file.


DoobesURU:
OK, so not *completely* finished...still needs the particle emitter for the calendar spark.

PR #22, opened 2021-02-27
Add Serene to MOULa

DoobesURU:
Adds PRP, LOC and OGG files for the Age of Serene for MOULa.


DoobesURU:
This version will not have the snow, and due to crashing, some more branches are bare. We can add back the snow when particles are added to Korman. In the meantime, the rest of the Age looks 1 to 1 and a Chiso Preniv Book has been added on an SDL state.

Hoikas:
Looks like Serene.loc is empty.

DoobesURU:
> Looks like Serene.loc is empty.

Odd...I wonder why it didn't push along with the other files. It looks OK on my end. :/

PR #24, opened 2021-03-18
VeeTsah LOC update

DoobesURU:
Adds Enders to the Veelay journals.


Hoikas (commented):
It looks like this sorts Enders by his IRL name instead of by his IC name. I expected him to be in the Es, but he was in the Ps.

DoobesURU:
> 
> 
> It looks like this sorts Enders by his IRL name instead of by his IC name. I expected him to be in the Es, but he was in the Ps.

Hmm...good point. I imagine most knew him by his avatar name. I'll change it.

PR #25, opened 2021-03-21
Add Vothol Gallery to MOULa

DoobesURU:
Adds PRPs and one SFX file for Vothol Gallery. It also changes Chiso slightly by rearranging the (future) Fehnir's House book and the Vothol book to avoid an empty pedestal in between.


Hoikas (commented):
Should 148f67448aff6b867301dd3cb72699a4b71c5f28 have been pushed to #24?

PR #28, opened 2021-04-06
Fix Neighborhood clock globe animation.

Hoikas:
[nbClockGlobe_(Entire Animation)_anim_0] Added a third keyframe to the clock rotation animation to cause the clock to actually rotate. Also converted to uncompressed quat keyframes due to lossy compression breaking the animation.


Hoikas:
Pulling for inclusion in Gehn.28 Beta 3.

Hoikas:
Validated on Minkata Shard. Merging...

PR #31, opened 2021-04-15
Fix clock globe turning backwards.

Hoikas:
Per observations in discord, the Neighborhood clock was inaccurate approximately four hours ago. The cause was part of the animation playing in reverse.

- [nbClockGlobe_(Entire Animation)_anim_0] Add two more keyframes to the
  animation to ensure the globe turns in the same direction at all
  times.


Hoikas:
I've updated this to fix the half pahrtahvotee offset issue reported in GitHub. Also, updated the resolution of the numbers in the lens and replaced the incorrect reflection of the classroom doors with a proper cube map rendered in-game of the plaza.

Screenshot:
Screenshot 2021-04-16 22 50 09

Hoikas:
Merging due to MOULa acceptance

Issue #18, opened 2021-01-14
Occluded transparencies not rendering


Grogyan:
Had a discussion in many, many years ago about the bug where when you are in 3rd person, and move your mouse over your avatar, if the texture in front of the avatar has an alpha transparency or has several alpha transparent textures, the rendering engine will make the area where the texture, behind it, invisible.

Particularly noticeable in the pods near the windows, but also on the Minkata cage.

Recently chatted with @dpogue on the Cyan Discord, and suggested that this issue be raised, even though it may not be possible to be fixed, as per the discussion over 10 years ago

Hoikas:
![image](https://user-images.githubusercontent.com/714455/127092403-c208795a-91d1-4f62-89c0-02c04949cc73.png)

PR #32, opened 2021-04-21
Fix render issues with user-created GZ Markers

dpogue:
This is accomplished by changing the RenderLevel on the DSpans from `0x40000000` (BlendRendMajorLevel) to `0x80000001` (LateRendMajorLevel + 1).

Partly addresses GitHub issue #18.

Screenshot from 2021-04-20 23-15-48

### DIFF
--- a/compiled/dat/GlobalMarkers_District_Markers.prp
+++ b/compiled/dat/GlobalMarkers_District_Markers.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:866778fa9352d8879de9d653fd997dd7dfa7d28d8e3af8c555295b24f79aefa8
-size 34343
+oid sha256:e0b9b1410898f27138ff0dc572d74248262a1e202ae1a2a1b7709a5454d6478e
+size 34341


Hoikas:
Have you checked standing in front of the marker and hovering the cursor over the avatar? IME, that's one of the biggest sources of x-raying. I think @DoobesURU has been using the blend onto framebuffer pass instead of the late rend pass when I report that in his Ages. I don't recall the difference in those passes off the top of my head.

dpogue:
I don't think I tried that particular scenario. The other part mentioned in #18 is the cursor-hovered avatar which seems like it's a problem with the avatar data moreso than anything else. Why did they get rid of `kAvatarRendMinorLevel`? 🤷🏼‍♂️

dpogue:
The marker does **not** show through the semi-transparent avatar. I'm still sort of thinking that's an issue with the avatar rather than the marker though.

Screenshot from 2021-04-21 13-35-59

Hoikas:
Ok, so I am thinking that the marker may need to be in the blend onto framebuf render level.

> Why did they get rid of kAvatarRendMinorLevel?

They didn't:
```c++
        if (val)
        {
            spans->SetNativeProperty(plDrawable::kPropSortSpans, true);
            spans->SetRenderLevel(plRenderLevel(plRenderLevel::kBlendRendMajorLevel, plRenderLevel::kDefRendMinorLevel));
        }
        else
        {
            spans->SetNativeProperty(plDrawable::kPropSortSpans, false);
            spans->SetRenderLevel(plRenderLevel(0, plRenderLevel::kAvatarRendMinorLevel));          
        }
```
(where `val` means: is the avatar opaque?)

Unless the kPropSortSpans logic is inverted...

dpogue:
If I didn't use `0x80000001` (even trying just `0x80000000`) the Marker would have alpha ringing when shown in front of the Relto fog plane.

Grogyan:
Very keen to see this fixed

Hoikas:
While this is not a completely correct fix, it is still a clear improvement over the previous state. Merging.

PR #34, opened 2021-05-06
Fix flat convex hull in the Neighborhood that caused crashes.

Hoikas:
- [VidCamScreensaver] Changed to kExplicitBounds


Hoikas:
Accepted to MOULa. Merging.

PR #39, opened 2021-05-14
Adds Kadish Butterflies

Hazado:
PRP provided by Siruis - http://forums.openuru.org/viewtopic.php?p=10053#p10053
Opened in PRPShop and saved as Moula PRP

No other changes made

Tested in game and seems like it worked.
Someone with CC Kadish Vault will need to do a comparison, as I dont have access to it at the moment


Hoikas:
Could you amend this commit to include this line at the bottom of the commit message:

`Co-authored-by: EhrenCG <EhrenCG@hotmail.com>`

This will make sure that we get the attribution right.

Hoikas:
Also, this appears to be a binary file instead of an LFS pointer

Hazado:
Should be fixed now

Hoikas:
Depends on H-uru/Plasma#914

Hoikas:
I've pushed a change to this PR that collapses the contents of the PRP provided in the original commit into Kadish_District_kdshYeeshaVaultIntr.prp.

PR #40, opened 2021-05-20
Fix for Yeesha Page 2 missing SDL hiding code

Hazado:
Working with Calum, noticed that you couldn't hide the Yeesha Page inside the Gahreesen Prison room.
After investigating, figured out the reason was no python code had been done for it.

This fixes the issue.


EhrenCG:
> 
> 
> Working with Calum, noticed that you couldn't hide the Yeesha Page inside the Gahreesen Prison room.
> After investigating, figured out the reason was no python code had been done for it.
> 
> This fixes the issue.

What about the fact that its variable "grsnYeeshaPage02Vis" appears twice in the SDL file?  Once grouped with "grsnYeeshaPage05Vis" under "# Boolean variables" and a second time under "# prison variables"?  That should probably be fixed too...

Hazado:
i agree, but im not sure what effect that will have on the vault database.

Deledrius:
Yikes, [looks like that's been the case](https://github.com/H-uru/Plasma/blob/master/Scripts/SDL/Garrison.sdl) since the first version in the repository.  That should probably get cleaned up for sanity's sake, but I'm not sure what the best way to do so is (aside from simply bumping the SDL desc version and removing one) as I'm not aware of what kind of undefined behavior this might be causing.  We'll probably need to do some testing around it.

Hoikas (approved):

Hoikas:
What an amusing oversight. I think we should probably ignore the duplicated SDL entry for now - that will be a bit more tricky. Thanks for the fix!

PR #41, opened 2021-05-22
Add UU Era Dance

Hazado:
Animations are added to be used globally.
I've tested the wall climbing animations in Gahreesen with no issue.
/dance2 works male and female

Translations done using google translate

Steps for converting CC Animation to Moula Animation
1. use libhsplasma prcdc.exe program to convert to XML file (prcdc.exe "location/of/cc/animation.prp" -o "location/to/save/xml/animation.xml")
2. open xml, rename the ageName to whatever age you want to use it in
3. change the location tags to be a new number (Location="-2;76" -> Location="-1;386")
4. change the plSceneNode's plKey Name to match the name of age (<plKey Name="Gahreesen_District_FemaleWallClimbDismountDown" -> <plKey Name="GlobalAnimations_District_FemaleWallClimbDismountDown")
5. Male Animations only - Duplicate the Left and Right UpperArms and rename one "UpperArm1" and the other "UpperArm2"
6. Male Animations only - remove all "Pinky" AppSets.
7. run libhsplasma prcc.exe on the xml using -v moul (prcc.exe -v moul "location/of/cc/animation.xml" -o "location/to/save/prp/animation.prp")

Co-authored-by: dpogue <darryl@dpogue.ca>


Hoikas:
This is a rather large changeset that is trying to do two things, from what I can tell:
- add the old UU era dance animation as an alternate
- move many Age specific animations to global

I think we need to separate those two concerns into separate changesets that can be more teasily digested.

Hazado:
Only PRPs are Old UU era Dance, Ball Push and Idle, Gahreesen Wall Climbing Animations.
The rest are all the XML files in the source directory, if we want to do further global animations.

Hazado:
Also https://github.com/H-uru/Plasma/pull/915 is tied to this one, to add in commands for /dance2

Hoikas:
Looks like there is still a conflict; you may have to rebase on top of master.

PR #45, opened 2021-05-29
Great Tree Pub Chiso page

Hazado:
Separated the Chiso section from the Pub.prp to its own page
Reverted all the Blender export changes made to both the Great Tree and Pub pages
Added back the Dynamic Cam Map work done by Doobes to the Pub
Changed audio of Imager back to reduced levels

https://github.com/H-uru/Plasma/pull/919 is needed for this to work correctly


Hazado:
Hmm, thought this was ready...need to export the changes and import into a fresh copy due to some rotation issues in the puzzle area

EhrenCG:
I'm confused about what's being done here and I can't figure out the thing for downloading large files, I hate to ask it, but is there another place to find these files?

Hazado:
> I'm confused about what's being done here and I can't figure out the thing for downloading large files, I hate to ask it, but is there another place to find these files?

https://www.dropbox.com/s/4f4ahpxfi41ddk5/GreatTreePub.7z?dl=0

I took the Chiso wing of the Pub, and made it its own page.
I then took a "pre-blender export" version of the pub and added back in the Dynamic Cam Mapping, which is what makes the floor reflective.
Now we have the Chiso version of the Great Tree Pub, without all the backend and lighting changes done.
If you compare to the current version, You wont fall through the floor behind the puzzle door. The lighting in the great tree area and the upper walkway is back to what it was originally. The secret door works again. There may be a few other things.

EhrenCG:
Hm, I think it's a little confusing calling the page just "Chiso".  Vothol uses the name "ChisoBook" for its book page, and in this case it's a room that could theoretically have two more books added to it later, not that I'm saying it's necessarily likely.  Maybe I'm just being weird since the name added the most to my confusion at first... kind of felt like you wanted to both separate and combine Chiso at the same time lol

So for the Chiso room itself the inside of the arch texture looks painted wrong, it looks like stripes rather than following the shape of the border itself.

Also, the responder modifiers for Er'cana "cRespLinkOutErca" and Ahnonay Cathedral "cRespLinkOutAhnonay" should have the fix mimicked to make sure the age display names show up correctly, one with the apostrophe and one with the space:

Changing <InstanceName value="AhnonayCathedral" /> to <InstanceName value="Ahnonay Cathedral" />
Changing <InstanceName value="Ercana" /> to <InstanceName value="Er&apos;cana" />

Hazado:
> So for the Chiso room itself the inside of the arch texture looks painted wrong, it looks like stripes rather than following the shape of the border itself.

> Changing InstanceName value="AhnonayCathedral" to InstanceName value="Ahnonay Cathedral"
> Changing InstanceName value="Ercana" to InstanceName value="Er'cana"

Fixed the above issues, Not sure what to call the section that Doobes made. Maybe Fan Annex?

EhrenCG:
> 
> 
> > So for the Chiso room itself the inside of the arch texture looks painted wrong, it looks like stripes rather than following the shape of the border itself.
> 
> > Changing InstanceName value="AhnonayCathedral" to InstanceName value="Ahnonay Cathedral"
> > Changing InstanceName value="Ercana" to InstanceName value="Er'cana"
> 
> Fixed the above issues, Not sure what to call the section that Doobes made. Maybe Fan Annex?

Looking at the link in point Doobes made for that room: "LinkInPointBookRm02"   I guess it could be simply be called "BookRoom02" or "FanBookRoom" or something along those lines?

Hoikas:
It looks like we have three open PRs that touch the Watcher's Pub PRP files... #45, #42, and #51. Unfortunately, GitHub is not able to merge PRP files for us - these changes would all overwrite each other on merge. @Hazado I would feel better if you worked with @DoobesURU to bring your changes into the GreatTreePub.blend file in this repository in one pull request. That way, we can more easily propagate the changes to new builds in the future.

GHGDelta:
Seems to me like the no camera blockers around the stairs and top section have had their collision 'modifier?' removed.  I used to climb those and get above pillars; I mean we must keep the cameras as intended.

Hazado:
Closing until a working blend file can be created

PR #46, opened 2021-06-03
Myst Online Intangible Assets

Hoikas:
Similar to the original, private drop from a few years back except licensed CC-BY-NA-SA 4.0. I took the liberty of omitting the client log directories and the "General Uru Info" directory containing the obsolete avatar sources. If the obsolete avatar sources are desired, I think we need to come up with a way to indicate its obsolescence to avoid mix-match problems with what's already in the repo.

Please double check this to make sure I did not accidentally commit binary objects instead of LFS pointers.


.gitattributes
```diff
@@ -6,23 +6,31 @@
...
 *.pxf filter=lfs diff=lfs merge=lfs -text
 *.tga filter=lfs diff=lfs merge=lfs -text
+*.tif filter=lfs diff=lfs merge=lfs -text
+*.ttf filter=lfs diff=lfs merge=lfs -text
 *.tron filter=lfs diff=lfs merge=lfs -text
 *.wav filter=lfs diff=lfs merge=lfs -text
 *.webm filter=lfs diff=lfs merge=lfs -text
+*.wps filter=lfs diff=lfs merge=lfs -text
 *.xcf filter=lfs diff=lfs merge=lfs -text
+*.zip filter=lfs diff=lfs merge=lfs -text
```
zrax, line 36: `*.xls` and `*.vsd` are binary files and should probably be added to this list

Hoikas:
I have pushed a new revision with all file extensions in the drop converted to lowercase. This should fix issues with files not being properly managed by LFS on non-Windows platforms.

dpogue (approved):

PR #47, opened 2021-06-05
Add GlobalClothing_Fan02 (Mysterium 2021 shirts)

Hoikas:
This adds male and female long and short sleeve T-Shirts for Mysterium 2021. The request was for these shirts to be used as a reward for completing some task in MOULa. Therefore they do not appear in the closet by default. To add them for all players on the shard, modify the Personal SDL's `GlobalRewardClothing` SDL to include both `SSMysterium2021` and `LSMysterium2021` (eg `SSMysterium2021;LSMysterium2021`). When awarding these items to specific players, use the `RewardClothing` SDL variable so as to not conflict with the global setting. The clothing will be added to the closet when the player visits Relto while these vault settings are active.

Screenshot 2021-06-05 00 07 40
Screenshot 2021-06-05 00 07 54
Screenshot 2021-06-05 00 10 13
Screenshot 2021-06-05 00 10 17

These PRPs are generated from the JSON source by the script CreateClothingPRP.py, originally by @Deledrius. The specific fork used to generate these files can be found at https://github.com/Hoikas/moul-utils. In the future, we should prefer for any new manually added clothing to be introduced using these JSON files instead of manually hacked into new PRPs.


Hoikas:
Bonus: Gay Zero Pride shirts from @dpogue.

Screenshot 2021-06-05 18 06 27
Screenshot 2021-06-05 16 40 13

Hoikas:
I have renamed the Mysterium shirts because per @tallpgr the previous SDL string exceeded the maximum allowable length. OP has been updated with new names.

Hoikas:
Accepted to MOULa. Merging.

PR #54, opened 2021-07-04
Fix text scaling in the release notes dialog.

Hoikas:
FHD comparison shot

The stretching was much worse on ultrawide:
Screenshot 2021-07-03 19 25 52


Hoikas:

PR #56, opened 2021-07-05
Fix solid bendy bridge in Teledahn.

Hoikas:
[PXPhysical] Bendy Bridge
Changed the member group from kGroupDynamic to kGroupStatic. The previous setting is for kickables (which this bridge is not). PhysX does not allow for non-kinematic dynamic bodies to have triangle mesh bounds. Because of the improper member group, we were forcing PhysX to treat the triangle mesh bounds as a convex hull, making the whole bridge solid.

Or, as Simulation.0.log says:
`WARNING: 'BendyBridge' is a dynamic triangle mesh; this is not supported in PhysX 4... forcing to convex hull, sorry.`


Hazado (approved):
Bendy Bridge is now appropriately bendy.

Tested jumping successfully and was able to walk on it like normal.

Deledrius (approved):

PR #29, opened 2021-04-06
Er'cana Collision Fixes

Hoikas:
This fixes two different Er'cana collision issues reported by Acorn and calumTraveller on discord.

- [PlantProxy27] Adjusted collision such that the avatar can walk or
  jump between it and PlantProxy26 without getting stuck between two
  parallel vertical planes.
- [RopeLadder03_TOP01] Extended the collision of the upper ladder region
  to ensure players cannot walk up to the edge and drop down,
  potentially getting stuck inside of the ladder's proxy collision.

PR #57, opened 2021-07-06
Er'cana Fix Rollup

Hoikas:
Requires H-uru/Plasma#954

Per @dgelessus
> * Have the harvester docked at the facility, with the little cart moved up/inside the facility
> * Board the harvester without moving the cart back down (easiest by Reltoing out and returning via the original link-in point)
> * Set the harvester to backwards direction
> * Push the call button for the little cart (on the harvester control panel)
> * Before the little cart arrives at the harvester, start the harvester so it moves away from the facility without the cart
> Now once the little cart arrives at the bottom, both the cart and the harvester lower their wooden bridges. Normally the cart should now be connected to the harvester, but it isn't, because you moved the harvester away first. Now you can simply walk off the wooden bridges on the harvester and you will fall through the world. The situation fixes itself once you send the harvester back to the facility and have it connect to the cart again.

This fixes:
- The bridges only lower if the harvester and car are stopped in the same location, preventing a trivial fall-from-world.
- The car no longer warps itself back to the beginning if you stop the harvester while the car is detached at the base of the factory.
- The car no longer warps itself to the bottom of the factory if you stop the harvester while the car is moving down from the silo to the base of the factory.

### DIFF
--- a/compiled/dat/Ercana_District_Canyon.prp
+++ b/compiled/dat/Ercana_District_Canyon.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:c81ef52306edb27bb3cbdc7d1804828c67bb036258f62aa35b744a5a8b561ec3
-size 4831759
+oid sha256:6191e52ac3e30a96465db37e699bcb7df42eedbb7898c8ffb1182ca777886781
+size 4626769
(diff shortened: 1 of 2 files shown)


Hoikas:
Conflicts with #29.

Hoikas:
All changes from #29 have been merged into the new PRPs from this PR to prevent conflicts.

Hoikas:

PR #58, opened 2021-07-06
Fix Neighborhood fire marble SFX

Hoikas:
~~WIP: This currently only fixes the four marbles in the entry area. Those marbles were already halfway set up for the sounds, the objects were just connected (very) incorrectly. The other three lack sounds entirely so need some more work.~~

This is a script generated fix for the Neighborhood firemarble kickable sound effects.

### DIFF
--- a/compiled/dat/Neighborhood_District_nb01.prp
+++ b/compiled/dat/Neighborhood_District_nb01.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:80997faf03a948704fe9359c3829af11a34a7e9adfe4845479504ff0b079d2d6
-size 7131501
+oid sha256:302a670d6bd694721d99c1cb2de0f7da731368dc4efdc97fb3708f93f09b3cd8
+size 7132909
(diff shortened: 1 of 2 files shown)


Hazado (approved):

EhrenCG:
Adding sounds from scratch to the other three marbles?
Let's see that's new objects of: 6 plSceneObjects, 6 plAudioInterfaces, 6 plWinAudibles, 6 plCoordinateInterfaces, 6 plRandomSoundMods, 27 freaking plWin32StaticSounds, and 3 plPhysicalSndGroups.  Since the sound buffers would already be there that's a total of 60 new objects along with needing to edit for the marbles: 3 preexisting plCoordinateInterfaces and 3 preexisting plGenericPhysicals for a 66 object job I guess you could call it.
Adding sounds manually is a long job...

Hoikas:
Yeah, I'm not sure I want to do it manually. I might script it or something. It's hard enough to hook these stupid things up in a GUI, let alone by manually hacking out the PRP objects.

Hoikas:
I reverted the manual pass at fixing the marbles and ran the [scripted conversion](https://github.com/Hoikas/moul-utils/blob/master/CreateKickableSounds.py) against the original nb01.prp file. This is a good thing because the results are completely reproducible and some confusing object naming is cleaned up. For example, previously, MarblePhy01 used SfxMarble-(Hit|Roll)Emit03 instead of SfxMarble-(Hit|Roll)Emit01 as you would expect. Running the script should be idempotent, so we're safe to modify the JSON file and re-run the script in the future.

This is now ready for prime-time.

Deledrius:
Marable?

Hoikas:
Fixed typo and missing newline at EOF.

Hoikas:

PR #62, opened 2021-07-15
Flip all Cyan LOC files through plLocalizationEditor.

Hoikas:
This round-trips the entire Cyan localization database through the plLocalizationEditor built from H-uru/Plasma#958. This means that:
- the XML files are re-encoded from UTF-16 to UTF-8 and therefore should load faster due to less run-time re-encoding
- journals should be written out as CDATA with the esHTML data visible in the clear


Hoikas:
Files have been re-flipped to fix parameter issue described in the H-uru/Plasma PR.

PR #66, opened 2021-07-22
Kadish Elevator fix

Hazado:
Should fix the collision issues noticed by testers on Minkata shards.
Testers were able to glitch through collisions on the front part of the elevator when it was rising up

[Scene Object - Elevator Wall] Coordinate Interface removed.
[Coordinate Interface - SubWorld_New] Scene Object "Elevator Wall" removed from Children
[Generic Physical - Elevator Wall] Bottom face removed. Angled and moved to bottom section of elevator shaft.


Hoikas:

PR #69, opened 2021-07-26
Fixes the Bahro Stone Linking Panel

Hazado:
Linking panel for bahro stones only have the right half of them actually usable.
This was caused by the change from the CC bahro share symbol to the Yeesha Share symbol.


Hoikas:
Why not just replace the too-small mipmap?

Hazado:
Im not sure if it is used in other places.

Deledrius:
Can you share a before and after for the AI image here?

Alderem:
Ideally, description should say _what about the bahro stone linking panel is being fixed_.  😅

Hazado:
Original
image
Upscale
image

Hoikas:
The image shouldn't be reused anywhere; let's go ahead and replace the old one.

Hazado:
Old one replaced instead. Force pushed.

Hoikas (approved):
Works great, thanks for the long-needed fix!

Deledrius:
Thanks.  The visual difference is barely discernible and produced no negative effects on the artistic aspects of the icon, so I'm happy to accept it here. :)

PR #68, opened 2021-07-25
Add Veelay Yeesha Page 26

DoobesURU:
Updated PRPs to add Yeesha Page 26 to Veelay Tsahvahn.

PR #71, opened 2021-08-21
Round firemarbles

Hazado:
Changed the firemarbles in Relto, Neighborhood and Kirel to be actually round.


Hoikas:
Conflicts with #68

Hoikas:
Based on discussion on discord, this is currently blocking on visual issues with the firemarbles. The ViewFaceModifiers seem to rotate in the "opposite" direction of the physical movement with this change applied. While the marbles tend to roll on a more realistic way with this change, the VFM problem is rather disorienting, so we should not accept until this is corrected.

ZarothYe:
I don't fully understand what it would take for the VFM issue to be resolved, but this PR will need to be rebased at the very least since there have been Neighborhood and Relto changes in the past 10 months.  I imagine it will need to be rebased yet again after all the pending Relto page PRs, too. Or, if it's not likely to be able to be completed any time soon, perhaps this needs to be closed or moved to draft state until someone can pick this back up.

PR #72, opened 2021-08-24
Enlarged the black mask of the telescope for ultra-widescreen resolutions

Hazado:
Before
image

After
image

### DIFF
--- a/compiled/dat/GUI_District_telescope.prp
+++ b/compiled/dat/GUI_District_telescope.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:63868d6235bf8ac1af75f993df6825a957940502b497ff2e58f6bd00a2afb709
-size 352989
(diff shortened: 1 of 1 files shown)


Hoikas (approved):

Deledrius:
What's the source of the image used in the new frame?

Hazado:
Thats default image, not sure why the first doesn't have it, was taken from Cyan Chat, #moula-tech-support.
I dont have a ultra-widescreen so I just used the reported image for the before.

Deledrius:
Ah, might be PotS vs MOULa.  I can't double-check a the moment.

Hazado:
Ahh I see the issue, they made all the color vertex to invisible basically.
I will fix this so its the same.

Hazado:
Not at widescreen, but here is the fixed texture
image

Hoikas (approved):

Deledrius (approved):

PR #75, opened 2021-09-03
Upscale a rug and blanket in Teledahn and Baron City Office

Hazado:
Was pointed out that the rug and blanket textures in these locations was rather low resolution compared to the rest of the textures around it.

Blanket is on the bed in Teledahn Office.
Rug is located in both Teledahn and Baron City Office.

The rug was recreated due to the excellent work of Calum Traveler, Tweek and Yali.
Blanket was upscaled using AI upscaler.

Blanket in teledahn 3rd floor office was moved to a new drawable span to have it use alpha.
Before it would have a black box around the edges.

Baron City Office
image
After
image

Teledahn
image
After
image


Hazado:
Oops, thought this was merged...

Hoikas:
I don't really remember where we were on this - ping @Deledrius

PR #77, opened 2021-09-28
Additional Graphical Fixes for Ercana

Hazado:
The warning light on the harvester console was not lit correctly.
Old 
image
New 
image

The harvester Railing had a few stretched and z-fighting textures and a few sections that the normals were flipped. The end caps for the railing near the ladder were also missing.
Old
image
image
image
New
image
image
image


Deledrius:
Looks good!

PR #78, opened 2021-09-29
Ayoheek message change to reference Gahreesen

Hazado:
Original references Garrison, but the name is Gahreesen now

Translations were done with Google Translate.

### DIFF
--- a/compiled/dat/HeekFrench.loc
+++ b/compiled/dat/HeekFrench.loc
@@ -3,7 +3,7 @@
 	<age name="Heek">
 		<set name="Messages">
 			<element name="NoKI">
-				<translation language="French">Vous devez posséder un KI pour jouer à Ayoheek. Explorez l'Âge Gahreesen et ramassez-en un.</translation>
+				<translation language="French">Vous devez posséder un KI pour jouer à Ahyoheek. Explorez l'Âge Gahreesen et ramassez-en un.</translation>
 			</element>
 			<element name="Rank">
 				<translation language="French">Vous êtes actuellement au rang %1s avec %2s point !</translation>
@@ -12,6 +12,7 @@
 				<translation language="French">Vous êtes actuellement au rang %1s avec %2s points !</translation>
 			</element>
 			<element name="SinglePlayerWarn">
+				<translation language="French">Vous êtes actuellement le seul à cette table Ahyoheek, deux personnes ou plus sont nécessaires pour jouer.</translation>
 			</element>
 			<element name="Welcome">
 				<translation language="French">Veuillez accueillir %1s, un joueur de rang %2s avec %3s point !</translation>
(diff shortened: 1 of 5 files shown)


compiled/dat/HeekGerman.loc
```diff
@@ -3,7 +3,7 @@
 	<age name="Heek">
 		<set name="Messages">
 			<element name="NoKI">
-				<translation language="German">Sie benötigen eine KI, um Ayoheek spielen zu können. Gehen Sie in die Garnisonswelt und holen Sie sich eine KI.</translation>
+				<translation language="German">Sie benötigen eine KI, um Ayoheek spielen zu können. Gehen Sie in die Gahreesen-Zeitalter und holen Sie sich eine KI.</translation>
```
Hoikas, line 6: It looks like the rest of the localization uses [Welt[en]](https://en.langenscheidt.com/german-english/search?term=welten&q_cat=%2Fgerman-english%2F) to mean Age[es]

So this should probably be `Gahreesenwelt` or `Gahreesen-Welt`. I don't speak German, so I don't know which one is actually correct.

Hazado: When I ran it through google translate it translated that as world instead of age. I dont speak german either. Changed it to "Age" and it came out as above

Hoikas: Yeah, I think in absence of a proper translation, we should prefer to be consistent with what's already there. So, I guess that means we want `Gahreesenwelt`.

compiled/dat/HeekFrench.loc
```diff
@@ -12,6 +12,7 @@
 				<translation language="French">Vous êtes actuellement au rang %1s avec %2s points !</translation>
 			</element>
 			<element name="SinglePlayerWarn">
+				<translation language="French">Vous êtes actuellement le seul à cette table Ahyoheek, deux personnes ou plus sont nécessaires pour jouer.</translation>
```
Hoikas, line 15: Was this translation done by a human? I'd prefer to avoid adding machine translations.

compiled/dat/HeekSpanish.loc
```diff
@@ -12,6 +12,7 @@
 				<translation language="Spanish">¡Ahora estás en el rango de %1s con %2s puntos!</translation>
 			</element>
 			<element name="SinglePlayerWarn">
+				<translation language="Spanish">Actualmente eres el único en esta mesa Ahyoheek, se requieren dos o más personas para jugar.</translation>
```
Hoikas, line 15: Same as above.

compiled/dat/HeekFrench.loc
dpogue: FWIW, this translation reads accurate to me

Hoikas (approved):

PR #79, opened 2021-10-02
Updates to GoMePubNew, VeeTsah, and Chiso Preniv

DoobesURU:
Updates the GoMePubNew, VeeTsah, and ChisoPreniv PRPs using the latest Korman exports. Chiso is also getting a new LOC file with a filename that better conforms to MOULa's naming scheme (plus one new entry for the Painter's Story (see changelog for Chiso below).

Also, I took the opportunity to make a few minor changes in each Age:

GoMePubNew:

* Adjusted menorah flames to show/hide with only one integer SDL state instead of 8 bool states (see PR #996 on Huru/Plasma) .
* Fixed the broken sitting mods in the corner alcoves.
* Added and removed a few items on the bulletin boards.
* A few adjustments in collision to keep Eddie and the Cones (sounds like a band name!) out of trouble. May still need more down the line as I had trouble recalling all the places they've been stuck.

VeeTsah:

* More adjustments to the ocean waveset, its mesh, and the material to (hopefully) improve its overall look.

ChisoPreniv:

* Rearranged the readable explorer books onto the shelves. Made them slightly brighter than the other books to make them more noticeable.
* Added the Tiam Book for future use and moved the Fehnir's House Book down the line again.
* Made god rays non-shadeless and emissive instead.
* Redid lighting on the shelves and a few other things as it was too dark on first export.
* Added new "Painter" story book from explorer DapperAndy.


Hoikas:
Standing inside the column

image

Hoikas:
Bad book! Bad!

image

Hoikas:
The only other thing I notice on a quick run-though is that the cones in the GoMePubNew still need kickable sounds. I wonder if some of the other kickables (eg the pumpkins) need some as well? At some point, now that Korman is no longer doing naughty things with lightmap UVs, I'd like to see if we could make some of the lightmaps smaller to reduce the size of some of these Texture PRPs.

Hoikas:
Does this supersede and close #48?

DoobesURU:
> Does this supersede and close #48?

It should, yes.

Hoikas:
Sorry, I closed the wrong pr. D'oh!

Hoikas (approved):

PR #83, opened 2021-10-07
Adds audio to Hood and Kirel Auditorium Podiums

Hazado:
This adds the audio button from CC inside the Hood and Kirel Auditorium location.

New Audio was also added to the Kirel podium which was recorded by Tyion
Original audio from CC was added back.


Hazado:
Fixed the Scene Nodes references.
Fixed the Null Soft volumes.
Changed the WinAudible to only be atteched to one emitter and use the index for each seperate audio.

Force pushed changes.

PR #84, opened 2021-10-19
Fix Quab footsteps

dpogue:
This ensures that the Armature Effects object for the Quab footsteps has the correct RandomSoundMod set up for the "Stone" surface, not just the "Dirt" surface.


dpogue:
I've updated this to include only the addition of the correct RandomSoundMod (done in PrpShop with PRC) and not any of the 3DS Max sources or re-exporting.

Hazado (approved):

Hoikas (approved):

PR #87, opened 2021-10-27
Restore missing footnote in King Kerath notebook

dgelessus:
Based on [a pre-release version of the King Kerath notebook](https://archive.guildofarchivists.org/wiki/Reference:DRC_research_notebooks/King_Kerath/2002_DPWR_transcription) that was preserved by the Guild of Archivists (formerly DPWR), the footnotes were slightly changed at some point during development (before the release of ABM). In particular, the first asterisk (next to a Kerath quote) was removed from the text, and the third footnote entry (referencing a Mee-Dis War book) was removed from the footnotes.

It appears that these changes were accidental. The removed asterisk doesn't match the removed footnote, and the remaining asterisks and footnotes no longer match up very well with each other. This is most noticeable with the "Taken from Kerath's public speech" footnote: in the pre-release version, it matches up with a Kerath quote, but in the release version, it corresponds to a paragraph about D'ni prophets that doesn't contain anything said by Kerath.

This commit re-adds the previously removed asterisk and Mee-Dis War footnote, which restores the footnotes to the more consistent pre-release state. The French and German translations are updated accordingly, with the translation of the footnote copied from other notebooks. The Italian and Spanish translations are *not* updated in this commit, because they are incomplete and don't contain a translation of the Mee-Dis War footnote anywhere.

I'm going to ping the Lore Team on the OpenUru Discord, so that they can look over this and check my reasoning, so that I don't rewrite history in a bad way.

### DIFF
--- a/compiled/dat/CityEnglish.loc
+++ b/compiled/dat/CityEnglish.loc
@@ -579,7 +579,7 @@ Kedri eventually passed away in 3903 at the age of 329, leaving the throne to hi
 
 His mother had raised him to follow the teachings of Gish and by the time he took the throne he was believed to be a whole-hearted believer in the Followers. Because of that, and the experience of watching his father interact with his advisors, Kerath had decided from an early age that a King was no longer the correct way that D'ni should be led. At least, he argued, not until the true Great King would come. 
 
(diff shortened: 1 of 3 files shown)


dgelessus:
> I'm going to ping the Lore Team on the OpenUru Discord, so that they can look over this and check my reasoning, so that I don't rewrite history in a bad way.

Alahmnat said on the OpenUru Discord that this is okay: https://discord.com/channels/282045216221822978/890674263210737665/903900275654406145

Hoikas (approved):

PR #89, opened 2021-11-01
Replace non-Latin-1 characters in .loc files

dgelessus:
Uru's fonts are limited to Latin-1 characters. If any characters outside that range are used in in-game texts, the characters are either not displayed at all or are replaced with a question mark symbol.

This PR replaces all non-Latin-1 characters in all .loc files with ASCII equivalents, so that they are displayed properly in-game. See commit message for details.


dgelessus:
By the way, GitHub isn't showing proper diffs for ChisoPreniv.loc and VeeTsah.loc, because those files are still encoded as UTF-16. #62 converted all the Cyan .loc files to UTF-8, but didn't touch the fan ones. Is there any particular reason why the fan .loc files should stay as UTF-16?

Hoikas:
The fan .loc files are all produced by Korman and will be converted to UTF-8 when a new export of those Ages is generated from Korman 0.12 (or higher). Right now, there is no way to import changes to the loc files into Korman, aside from copying and pasting said changes changes. Could you back out the changes to ChisoPreniv.loc and VeeTsah.loc for now and work with @DoobesURU to get those fixed in the source material?

Hoikas:
Hmm, the web interface is mad that I merged #87 before this one.

dgelessus:
That's because there was an actual merge conflict 🙂 I just did a rebase on top of the current master, so it should be happy now.

compiled/dat/GahreesenSpanish.loc
```diff
@@ -15,7 +15,7 @@ De acuerdo, una cosa parece completamente obvia; este sitio fue construido como
...
 
-Debo decir que Kodama tiene algunas teorías acerca de la Era, y que sigue sosteniendo hasta la fecha. Puede que hayan existido una especie de “fuerzas especiales” de una clase de Reparadores. Comenzaron algo más tarde, a mediados del 8.000. Convirtiéndose en algo parecido a un arma para la investigación y el desarrollo del Gremio de los Reparadores. No lo sé, aunque merece ser mencionado. Existen algunas menciones de dichos grupos que han sido encontradas por Kodama en otros documentos, o al menos que se hayan visto. 
+Debo decir que Kodama tiene algunas teorías acerca de la Era, y que sigue sosteniendo hasta la fecha. Puede que hayan existido una especie de "fuerzas especiales" de una clase de Reparadores. Comenzaron algo más tarde, a mediados del 8.000. Convirtiéndose en algo parecido a un arma para la investigación y el desarrollo del Gremio de los Reparadores. No lo sé, aunque merece ser mencionado. Existen algunas menciones de dichos grupos que han sido encontradas por Kodama en otros documentos, o al menos que se hayan visto.
```
Hoikas, line 18: This is might need to be left to another PR and a native speaker, but I believe that angle brackets should be used instead of quotation marks:

`«fuerzas especiales»`

Hoikas (approved):

compiled/dat/GahreesenSpanish.loc
dgelessus: The existing translations are a bit inconsistent about which quotes they use. For example, you can find French translations using either « guillemets » or "quotes" (and as far as I know, the first is standard in French). I was planning to look at that after this PR. I wanted to fix all the invisible punctuation first, without changing the quote style yet, to avoid mixing a technical change with a stylistic one.

Hoikas: That sounds like a good plan to me.

PR #91, opened 2021-11-02
Make quote styles consistent across all .loc files

dgelessus:
Especially the French and German translations were inconsistent with their quote styles. In both cases I've adjusted all quotes to match the standard style for the language, i. e. « guillemets » for French and "double quotes" for German.

@Hoikas also [suggested](https://github.com/H-uru/moul-assets/pull/89#discussion_r740558683) that the Spanish translation should also use «angle brackets», but wasn't sure as he's not a native speaker. I know nothing about Spanish myself, so I didn't make any changes to the Spanish quotes.


Hoikas:
@dgelessus Let us know when you have a chance to fix the conflicts.

dgelessus:
Oops, didn't notice that. Rebased to fix the conflict.

Hoikas:
Thanks for the fixes :tada:

PR #94, opened 2021-12-16
Spyroom Q4 Updates

DoobesURU:
Brings the spyroom 1 to 1 with the current version on MOULa.


Hoikas:
Deferring until blend is also available

Hoikas:
What's the status here? Were these PRPs deployed to MOULa?

DoobesURU:
> What's the status here? Were these PRPs deployed to MOULa?

Yup. These are actually direct copies from MOULa. Only minor change was adding the Aussie hat. And, of course, it's now a Korman export.

Hoikas:
OK, so we're just waiting for the blend to be added before merging, right?

DoobesURU:
> OK, so we're just waiting for the blend to be added before merging, right?

Yep. I'll be doing that shortly in a new commit.

Hoikas (approved):

PR #97, opened 2022-01-06
Add Battlebotsboy and Grandbad to Veelay Journals

DoobesURU:


Hoikas (approved):
Looks like this also converts the LOC to UTF-8, which is nice.

PR #98, opened 2022-01-06
Fixed Weight Tables for CC imported clothing

Hazado:
Due to badly transferred weight tables from the CC max files to MOULA max files, the CC clothing would glitch when seen from a distance on avatars. I've now fixed that issue.

Male03.max and Female.02.max
Weight Table for backpack 2 and foot06 corrected

Male04.max and Female.04.max
Weight Table for Atrus Legs, Catherine Legs and shoes corrected

(Fe)Male02
Re-exported and saved

(Fe)Male04
Re-exported and saved


Hazado:
I tested the fan clothing, with no breakage.
I can flip it through though if you want

Hoikas (approved):

PR #106, opened 2022-02-28
Fix nexus KI slot regions in City

Hazado:
Resolves H-uru/moul-assets#107

Nexus Book and Ki Slot regions had a number of issues.

Fixed so you can always click the left side and it will add the link to you nexus.
Clicking on book will open the book.


Hoikas (approved):
This is a clear improvement. Thanks for fixing this long-standing problem :+1:

PR #108, opened 2022-03-01
Change PRP for options menu GUI to include Enable Subtitles checkbox

ZarothYe:
Splitting out PR for options menu checkbox from PR for subtitle text files (#101)


Hoikas:
Blocking on:
- H-uru/Plasma#1063
- KI changes to be done in xKIGUI.max (instead of directly to the PRP files)

Hoikas (approved):

PR #109, opened 2022-03-01
Add single subtitle file for Zandi's Cleft intro as a testing sample

ZarothYe:
Adding single subtitle file (easy to test with new character) as a sample to test the new feature. I was promised bonus points by Adam:

> I think there would be bonus points for breaking up the .sub PR into a single "demonstration" subtitle PR to go along with the core functionality and a larger one with the bulk load of the .subs. That way we can merge the functionality itself rather quickly and take time glancing over the subtitles themselves.


Hoikas:
Should be merged concurrently with H-uru/Plasma#1063.

Hoikas (approved):

Hoikas:
:shipit:

PR #110, opened 2022-03-11
Remove triple-double-quote endings from Descent and Phil's Relto journals

ZarothYe:
Exactly what it says on the tin. :)


Hoikas (approved):
lol, someone mad a boom boom when they copied the text out of the Python source file.

PR #112, opened 2022-03-11
Fix GZ Beam in city

Hazado:
GZ beam animation end point is past its start point.
This give the impression that the GZ beam is skipping backwards when it gets to the arch.

Adjusted the end point so it better meets the start point.


Hoikas:
Did you intend to make changes to DrawableSpans city_harbor_00000000_0Spans or is this just noise from round-tripping through libHSPlasma?

Hazado:
noise from roundtripping through libhsplasma, texture objid changed for the city rocks

Hoikas:
Yep, the hsGMaterial `rock arms` changed from ObjID 1 to 12.

Hoikas (approved):

PR #116, opened 2022-03-25
Adds the cyan camera work done to create the trailer for Myst Online

Hazado:
This is the cleft camera flyby done in this trailer for Myst Online
https://www.youtube.com/watch?v=rLvkMJDrMAU

Clip of how I set up to trigger the camera and the flybys
https://www.youtube.com/watch?v=8An_EP3jVfk

This PRP will never load in any game due to the age file having the page set to be controlled by page loading.
Inside the age file Line 35
`Page=trailerCamPage,18,1`
needs to be changed to load the file
`Page=trailerCamPage,18`

### DIFF
--- a/compiled/dat/Cleft_District_trailerCamPage.prp
+++ b/compiled/dat/Cleft_District_trailerCamPage.prp
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:9a979af9e1b1577a5dfe0ac9f88da1bbe7f888e38bdb2740a1614ed9cbf7c913
+size 361370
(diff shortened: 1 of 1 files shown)


Deledrius:
Neat!

Hoikas:
@Deledrius Should we mainline this? I don't see any harm in taking it.

Deledrius:
Seems like a good idea to me!

Hoikas (commented):
I think that, to prevent surprises, we should remove the trigger region and instruct anyone who wants to use this to execute the console command `Logic.TriggerResponder runAnimCam` instead.

Hoikas:
Also, the next time your tweak this, it would be nice if the `runAnimCam` responder popped the trailer camera off the camera stack when the animation ends.

Hazado:
> I think that, to prevent surprises, we should remove the trigger region and instruct anyone who wants to use this to execute the console command `Logic.TriggerResponder runAnimCam` instead.

I added the trigger region because i couldnt figure out the commands to trigger it 😭

Hoikas:
Send another `plCameraMsg` when the animation finishes with *only* the `kRegionPushCamera` flag set. Be sure to *not* set `kEntering` or `kResponderTrigger`.

Hazado:
Okay, no more trigger region and regular camera is returned after animation is done.

Hoikas (approved):

PR #117, opened 2022-03-25
Add linking panels and new link point into Gahreesen

Hazado:
Added a Linking point inside the Gahreesen Training Control Center (place above the prison cells)
This linking point was talked about inside the Garrison Design doc (https://ia902800.us.archive.org/26/items/urudesigndocs/Garrison.pdf), Tower Desk on page 15.

It required adding a new subworld and animation that spins the metal disk in the center of the room.

image

Also added the linking panels for the Garrison Maintainer Nexi inside GUI_District_BkBookImages.prp


Hoikas:
Thanks for helping clear out some old PRs, @Hazado. I may want to take another look at this one in the near future - just need to get the BkBookImages PR up first.

Mystler:
With #289 resolved, I think this is now finally unblocked and can be revisited by having the sources for the new link panels in the texture sources and json file.

PR #119, opened 2022-04-10
Add /textcolor command and supporting text elements

ZarothYe:
~~No translations available yet for command, default arg value, status messages, or error/usage message. I need to reach out to some translators in the OU discord to get at least French and German, probably. I'm not 100% happy with the wording in the error/usage message in English, though, so I'd like to get some suggestions from y'all and smooth that out first.~~

Reworked error/usage message to a place where I am reasonably happy with it. French and German translations kindly provided by skyisblu and @dgelessus, respectively. Some Spanish color names based on old college Spanish knowledge and Google but I don't expect we'll get a native translator for the rest of the functionality in Spanish ~~or in Italian~~ any time soon. I was wrong about the Italian functionality; Italian has been kindly provided by aurora and Korovev in the OU discord.

Ready for review and merge whenever if there is no more feedback.

### DIFF
--- a/compiled/dat/GlobalSpanish.loc
+++ b/compiled/dat/GlobalSpanish.loc
@@ -11,13 +11,18 @@
 			<element name="Brown">
 				<translation language="Spanish">café</translation>
 			</element>
-			<element name="Cyan">cian</element>
+			<element name="Cyan">
+				<translation language="Spanish">cian</translation>
+			</element>
 			<element name="DarkBrown">
+				<translation language="Spanish">caféoscuro</translation>
 			</element>
 			<element name="DarkGreen">
 				<translation language="Spanish">verdeoscuro</translation>
 			</element>
-			<element name="DarkPurple">morado</element>
+			<element name="DarkPurple">
+				<translation language="Spanish">morado</translation>
+			</element>
 			<element name="Gray">
 				<translation language="Spanish">gris</translation>
 			</element>
@@ -27,8 +32,12 @@
 			<element name="Magenta">
 				<translation language="Spanish">magenta</translation>
 			</element>
-			<element name="Maroon">granate</element>
-			<element name="NavyBlue">azulmarino</element>
+			<element name="Maroon">
+				<translation language="Spanish">granate</translation>
+			</element>
+			<element name="NavyBlue">
+				<translation language="Spanish">azulmarino</translation>
+			</element>
 			<element name="Orange">
 				<translation language="Spanish">anaranjado</translation>
 			</element>
@@ -39,9 +48,14 @@
 				<translation language="Spanish">rojo</translation>
 			</element>
 			<element name="SlateBlue">
+				<translation language="Spanish">azulcielo</translation>
 			</element>
-			<element name="SteelBlue">verdeazulado</element>
-			<element name="Tan">caféclaro</element>
+			<element name="SteelBlue">
+				<translation language="Spanish">verdeazulado</translation>
+			</element>
+			<element name="Tan">
+				<translation language="Spanish">caféclaro</translation>
+			</element>
 			<element name="White">
 				<translation language="Spanish">blanco</translation>
 			</element>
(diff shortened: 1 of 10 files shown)


compiled/dat/KIEnglish.loc
```diff
@@ -139,6 +139,12 @@
 			<element name="TOPrompt">
 				<translation language="English">TO:</translation>
 			</element>
+			<element name="TextColorSetDefault">
+				<translation language="English">Text color now set to default color(s).</translation>
+			</element>
+			<element name="TextColorSetValue">
+				<translation language="English">Text color now set to %1s.</translation>
+			</element>
```
Hoikas, line 147: I think these should probably say "reset" instead of "now set."

ZarothYe: You think both of them should say "reset"? One is used with `/textcolor default` to reset to normal state, the other is used with `/textcolor {newVal}` to set to a new value, so that one feels a little strange to use "reset" to me.

Hoikas: Good point. I think just the default variant should say "reset".

compiled/dat/GlobalSpanish.loc
```diff
@@ -1,6 +1,60 @@
...
+			</element>
+			<element name="Cyan">
+			</element>
+			<element name="DarkBrown">
+			</element>
+			<element name="DarkGreen">
+				<translation language="Spanish">verdeoscuro</translation>
+			</element>
+			<element name="DarkPurple">
+			</element>
```
Hoikas, line 22:
```suggestion
			<element name="DarkPurple">morado</element>
```

As a non-native speaker, this was what seemed to match after looking at images of these purple-ish colors:
- violeta (what I learned as "purple")
- púrpura (this also means bruise)
- morado

compiled/dat/GlobalSpanish.loc
```diff
@@ -1,6 +1,60 @@
...
+			<element name="Pink">
+				<translation language="Spanish">rosa</translation>
+			</element>
+			<element name="Red">
+				<translation language="Spanish">rojo</translation>
+			</element>
+			<element name="SlateBlue">
+			</element>
+			<element name="SteelBlue">
+			</element>
```
Hoikas, line 48:
```suggestion
			<element name="SteelBlue">verdeazulado</element>
```

compiled/dat/GlobalSpanish.loc
```diff
@@ -1,6 +1,60 @@
...
+			<element name="Tan">
+			</element>
```
Hoikas, line 50:
```suggestion
			<element name="Tan">caféclaro</element>
```

Or `marrónclaro` with the caveat that this is probably Spain only.

compiled/dat/GlobalSpanish.loc
```diff
@@ -1,6 +1,60 @@
...
+				<translation language="Spanish">naranja</translation>
+			</element>
+			<element name="Pink">
+				<translation language="Spanish">rosa</translation>
+			</element>
+			<element name="Red">
+				<translation language="Spanish">rojo</translation>
+			</element>
+			<element name="SlateBlue">
+			</element>
```
Hoikas, line 46: This will be either `azulcielo` or `azulceleste` for "skyblue", but I'm not sure which one is better.

ZarothYe: I am inclined to lean toward `azulcielo` just based on the usage of `skyblue`, `bleuciel`, and `himmelsblau` in other translations, but that's not a particularly strong or compelling reason if someone else feels the other word is more fitting. It was requested that the valid colors be listed out in the usage message so it takes the guesswork out of using the command, and I think that makes sense. All that to say, whichever you pick will be fine, although without the rest of the Spanish status message translations the usability will be weird anyway.

Hoikas (approved):

PR #120, opened 2022-04-15
Add /cough and its emote string (plus translations)

ZarothYe:
German translation approved, French and Italian are just best guesses for now.


Hoikas:
Is this currently in use by the KI or are we also waiting on corresponding scripts changes?

ZarothYe:
> Is this currently in use by the KI or are we also waiting on corresponding scripts changes?

It was already listed in the scripts as an emote, but there were no entries in the LOCs to correspond for some reason. It is similar to the /blowkiss emote in that it gives an engine animation name that doesn't exist, though.

PR #124, opened 2022-04-23
Fix reference to stale DSpan object in Markers.

Hoikas:
This fixes a crash, presumably introduced by #32, caused by references to a stale DSpan object. The crash could be observed by creating a marker game an attempting to add a marker to it.


ZarothYe:
This fix appears successful, as I can now add markers to my game on TrollLand 🥳 
TLMarker

dpogue (approved):

PR #127, opened 2022-04-27
Remove spaces in Spanish emote commands and guilt-trip in quit confirmation

ZarothYe:
- Remove spaces in several emote command strings. Spaces in emotes are not acceptable. E.g., `/no sé` just does the `/no` "Shake Head" emote and a separate message afterwards just saying "<PlayerName>: sé". These were bad and confusing times.
- Remove duplicated emote commands `/nosé` and `/gracias`. Second instance of `/gracias` was replaced by `/agradecer` instead.
- Fixed the emote string for the `/agradecer` emote... because its message was in Italian. These were also bad and confusing times.
- Remove the `¡No te vayas por favor!` (roughly, "Please don't go!") part of the quit/logout confirmation message. Apparently a holdover from Choru, or so I am told. Anyway, none of the other languages get so emotional about a player quitting so it feels inappropriate to keep.

### DIFF
--- a/compiled/dat/KISpanish.loc
+++ b/compiled/dat/KISpanish.loc
@@ -147,7 +147,7 @@
 			<element name="AddBuddy">
 			</element>
 			<element name="Afk">
-				<translation language="Spanish">vuelvo en seguida</translation>
+				<translation language="Spanish">vuelvoenseguida</translation>
 			</element>
 			<element name="Amazed">
 			</element>
@@ -180,7 +180,7 @@
 				<translation language="Spanish">/amigos</translation>
 			</element>
 			<element name="ChatClearAll">
-				<translation language="Spanish">/autorizar chat</translation>
+				<translation language="Spanish">/autorizarchat</translation>
 			</element>
 			<element name="ChatNeighbors">
 				<translation language="Spanish">/sectores</translation>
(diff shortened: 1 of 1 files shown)


dpogue:
This is more of a general observation than something specific to this PR, but I noticed (both here and in English) that the commands are inconsistently prefixed with a slash. Is it meaningful whether the slash is present or not, or does the game not care? Should we attempt to make them consistent at some point in the future?

ZarothYe:
> This is more of a general observation than something specific to this PR, but I noticed (both here and in English) that the commands are inconsistently prefixed with a slash. Is it meaningful whether the slash is present or not, or does the game not care? Should we attempt to make them consistent at some point in the future?

It is meaningful. This is more to do with how the engine handles different types of commands. Emotes are never prefixed, and are checked last by the engine after stripping off the leading slash in the user's message. Other types of commands like chatlogging, the various "petition" commands (which seem not to ever be used), or adding or removing buddies or ignorees are always prefixed and checked exactly as-is. We probably could change the handling and add slashes to all the emote commands in the LOCs and such to make them consistent, but I'm not certain there's a lot of benefit at this point. Y'all's call, though.

dpogue:
Ahhh okay, that sounds like a discussion for another day

Hoikas (approved):

PR #128, opened 2022-04-28
Add Elonin

DoobesURU:
Adds the cavern location of Elonin by Dulcamara and maintained by ametist


Hoikas:
At this time, I am of the opinion that this contribution is not ready for mainline.

Hoikas:
There is something not quite right with the line endings of `Elonin.loc`. This PR consistently breaks my deployment scripts due to `Elonin.loc` always being "changed."

Hoikas:
Another deployment broken by Elonin.loc

Hoikas:
Another deployment broken by Elonin.loc

ametist365:
Can I do something to fix that at my end? I exported the age files with 'Localization Database(Compat Mode)' set.
I see above that 'there is something not quite right with the line endings in Elonin.loc'

ZarothYe:
IIRC, @Hoikas wondered at one point if Elonin was being exported using an older version of Korman and getting set with the wrong type of line endings. What version of korman are you using to export Elonin, @ametist365 ? I think it could also be an issue with Doobes's git settings for `auto.crlf` messing with line endings prior to upload, but I'm not certain. If you have the Notepad++ program installed (free feature-rich text editor), you can open the LOC file and view the line endings using **View > Show Symbol > Show End of Line** button and also change the line endings if needed using **Edit > EOL Conversion > Unix (LF)**. The line endings should be LF instead of CRLF for the H'uru repo.

@DoobesURU The changing of the CDATA usage to &gt; and &lt; that you have been trying should not make a difference, I think. The CDATA usage is just fine; the successful deployment of Elonin to TrollLand some months ago was with CDATA included. I strongly suspect it's the line endings instead.

ametist365:
I use the latest korman 0.13a and when I export I use 'Localization Database(Compat Mode)' per Adams advice. Today I have downloaded and installed Notepad++ and will do a change of the line endings if needed.

DoobesURU:
On advice from @ZarothYe and @Hoikas , closing this rather broken PR in favor of a fresh one.

PR #135, opened 2022-05-04
Minkata Soccerball issue

Hazado:
The Minkata Soccer ball with the OU client will be invisible until the ball is moved a bit.
This is a work around for that behavior and moves the ball up to 5 feet from 1 foot.
When the age is initialized it causes the soccer ball to fall and become visible.

This issue does not affect the H'uru client at all.
These changes cause no visible change to the soccer ball on H'uru client.

This PR is strictly to keep the two codebases in sync with each other.

### DIFF
--- a/compiled/dat/Minkata_District_minkCameras.prp
+++ b/compiled/dat/Minkata_District_minkCameras.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:1e6ae6b53f414b7ab6184b243f62011e713b0dcbd5f6ad422dd08bd0221cd6c3
-size 97551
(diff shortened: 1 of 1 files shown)


Hoikas:
Based on discussion in discord, this is not the correct fix even for the obsolete client. In this specific case, the ground is in a page that is loaded after the game loop runs at least once during the Age load while the kickable is in an on-by-default page. The correct fix for the obsolete client is to ensure that kickable objects do not begin falling while the Age is loading, as in commits [027e83036023bee3d182c2bd3266a3b09e11b06e](https://github.com/H-uru/Plasma/commit/027e83036023bee3d182c2bd3266a3b09e11b06e) and [bc76f4d760bd7f72288a59ab8b84fea47807612c](https://github.com/H-uru/Plasma/commit/bc76f4d760bd7f72288a59ab8b84fea47807612c).

Issue #125, opened 2022-04-23
Missing Tiam files?


Hoikas:
Per UruManifest

```
[2022-04-23 12:01:57,592] ERROR: Dependency file 'sfx/Tiam_gitter.ogg' (provided by: 'audio/Tiam_gitter.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,593] ERROR: Dependency file 'sfx/Tiam_gitter.ogg' (provided by: 'audio/Tiam_gitter.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,593] ERROR: Dependency file 'sfx/Tiam_Liddown.ogg' (provided by: 'audio/Tiam_Liddown.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,593] ERROR: Dependency file 'sfx/Tiam_Liddown.ogg' (provided by: 'audio/Tiam_Liddown.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,593] ERROR: Dependency file 'sfx/Tiam_Pling.ogg' (provided by: 'audio/Tiam_Pling.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,594] ERROR: Dependency file 'sfx/Tiam_Pling.ogg' (provided by: 'audio/Tiam_Pling.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,594] ERROR: Dependency file 'sfx/TiamWatLapping3.ogg' (provided by: 'audio/TiamWatLapping3.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,594] ERROR: Dependency file 'sfx/TiamWatLapping3.ogg' (provided by: 'audio/TiamWatLapping3.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,594] ERROR: Dependency file 'sfx/TiamSten3.ogg' (provided by: 'audio/TiamSten3.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,595] ERROR: Dependency file 'sfx/TiamSten3.ogg' (provided by: 'audio/TiamSten3.ogg') missing! Used by 'Tiam'.
[2022-04-23 12:01:57,595] ERROR: Dependency file 'sfx/Tiam_dblbuttnew.ogg' (provided by: 'audio/Tiam_dblbuttnew.ogg') missing! Used by 'Tiam'.
[...]

PR #140, opened 2022-05-12
Add Tiam SFX

DoobesURU:
Adds the missing Tiam sound effects to fix issue #125 .


ZarothYe (approved):
Verified these 16 files are the complete set of sfx files for Tiam that were put in place on MOULa Prod, and the file names and sizes match what is present in Prod as well.

Hoikas (approved):
Verified no additional errors are reported by deployment scripts.

Tds33:
Hi

PR #143, opened 2022-05-17
Adds Ocean Page

Hazado:
This adds the Ocean page to required files to have it work in Relto.

Still WIP as its missing the ripples around the islands.

There is no in-game spot to acquire this page yet. Still trying to decide where to put it.


Hazado:
image
image
Some screenshots

Hazado:
Ocean page does not have a physical in-game page yet.
Content is ready for use otherwise

Deledrius:
That's a really neat concept!  The shores look good on the small islands.

The water "texture" itself appears to be quite stretched out in the provided screenshots.  Is this an issue with the underlying mesh density?

Hazado:
> That's a really neat concept! The shores look good on the small islands.
> 
> The water "texture" itself appears to be quite stretched out in the provided screenshots. Is this an issue with the underlying mesh density?
> 
> Also: does this make the region off-island swimmable? If so, how far does this go?

Water texture is dynamic reflection of the sky, can only make it so good without destroying graphics cards and computers unfortunately.

Water for the small off-shore islands is too low to climb out on them.
The clock Island and one next to it you can climb onto.

It goes pretty far, ive got an invisible wall blocking from swimming too far out.

Deledrius:
> Water texture is dynamic reflection of the sky, can only make it so good without destroying graphics cards and computers unfortunately.

The choppiness and stretching aren't based on the sky itself, though, as I recall that's an element of the waveset object.  Is the mesh density not high enough near the shore, perhaps?

> Water for the small off-shore islands is too low to climb out on them. The clock Island and one next to it you can climb onto.

That sounds excellent. :) 

> It goes pretty far, ive got an invisible wall blocking from swimming too far out.

Hmm, that's unfortunate.  It's perhaps unavoidable with this addition, but that's precisely the reason the game had previously been designed very particularly to avoid invisible walls.  The closest thing is the "lost in the dust storm" of Minkata, and it's heavily disguised there.  There are no swimming areas that are not physically bounded specifically to avoid the "invisible wall" problem.  I wonder if something more interesting could be used here?  A distant ring of fog, similar to Minkata?  Or the entire body of water mysteriously dropping off a waterfall at the extreme edge of the ring?

It would be a good idea to maintain the existing rules of gameplay and overall design when adding new content, even something that's mostly cosmetic like this.  We risk a slippery slope of inconsistency if the extra effort is not made with our fan content additions.

dpogue:
> > It goes pretty far, ive got an invisible wall blocking from swimming too far out.
>
> Hmm, that's unfortunate. It's perhaps unavoidable with this addition, but that's precisely the reason the game had previously been designed very particularly to avoid invisible walls. The closest thing is the "lost in the dust storm" of Minkata, and it's heavily disguised there. There are no swimming areas that are not physically bounded specifically to avoid the "invisible wall" problem. I wonder if something more interesting could be used here? A distant ring of fog, similar to Minkata? Or the entire body of water mysteriously dropping off a waterfall at the extreme edge of the ring?

A current region that pushes back towards the main island, similarly to Ahnonay (before turning off the current)?

In any case, since this isn't accessible in MOULa right now, these sound like things that could be tweaked before it gets enabled, rather than merge blockers.

Deledrius:
> A current region that pushes back towards the main island, similarly to Ahnonay (before turning off the current)?

A current isn't a bad solution, as long as it takes effect fairly close (within visibility range) of the main island.  

If it's too far out it becomes hard to tell the avatar isn't moving without a frame of reference.  Adding more islands for visual anchoring would suggest to the player that there's something out there to explore (unlike the Minkata craters which are communicated fairly clearly to the player as being visual filler), and that's misleading and disappointing.

> since this isn't accessible in MOULa right now, these sound like things that could be tweaked before it gets enabled, rather than merge blockers.

I agree.

Hazado:
> 
> A current region that pushes back towards the main island, similarly to Ahnonay (before turning off the current)?
> 
> In any case, since this isn't accessible in MOULa right now, these sound like things that could be tweaked before it gets enabled, rather than merge blockers.

That was my intentional.
I ran out of time on this current release.
What is included on this pull request is what is currently on MOULA.

My plan currently for this page is to add the water that pushes back in a ring around the edge and not leave it at the invisible wall.

Deledrius:
Understood!  That sounds like a good plan. :)

PR #148, opened 2022-05-18
Adds Tiam Crystal page

Hazado:
Adds the objects and assets for the Tiam Crystal page.

Page will be found inside the Tiam Age.
Will be another PR for that.


ZarothYe:
I think I saw some screenshots for this page (and the other pages) in discord at one point. Can we add some here to give some idea of what the page graphic and results in Relto will be for the people who may not be following the discord? And for posterity since discord history is hard to trawl through? :)

Hazado:
image
image

Deledrius (changes requested):
This is an interesting idea for an addition.  It reminds me of the crystal trees in Ahnonay.

The texturing on these seems off though, as it doesn't match the lighting in the locations of Relto they are growing from.  It looks like a full-bright object sitting on Relto, but without any glow that would justify that appearance.  The UV mapping also looks a bit haphazard in places, and the texture itself is very simple and repetitive.  Compare it to [the crystal trees on Ahnonay](https://i.pinimg.com/originals/91/e9/7a/91e97abbc5890ae33ed668275a675d7a.jpg), and you can see there a variation in texture and lighting that makes them actually fit in at the place they've been planted.

I would suggest this needs some blended textures and more thought to the UV mapping before it's considered ready.  

Right now this is a good first pass at a cool new Relto page.

Hazado:
They are meant to match the Crystals from Tiam age. UV Mapping is the same as used in Tiam.
They do have a glow in-game.

Deledrius:
> They are meant to match the Crystals from Tiam age. UV Mapping is the same as used in Tiam. They do have a glow in-game.

Ah, so this should be blocking on Tiam as well.  They look unfinished and don't match the visual fidelity expected here in Relto, and either don't match Tiam either or Tiam itself needs additional work as a whole.

Deledrius:
> I think I saw some screenshots for this page (and the other pages) in discord at one point. Can we add some here to give some idea of what the page graphic and results in Relto will be for the people who may not be following the discord? And for posterity since discord history is hard to trawl through? :)

Indeed, I believe this review is the first I'm seeing of these additions.

Hoikas:
Thanks for the submission. Unfortunately, at this time, we are not accepting new Ages to this repository. If you are still interested in submitting your new Age, please submit it directly to the Shard(s) that you are interested in it appearing on. Thanks!

PR #160, opened 2022-06-04
Add localized audio tracks to AtrusIntro.webm

ZarothYe:
Added the other 4 audio tracks to AtrusIntro and added the correct language properties to them to ensure they play according to the game's selected language. File has increased from ~10MB to ~30MB. Original .bik file from CC was ~32MB, for reference.


Deledrius:
Awesome!  I was really hoping we'd be able to utilize this functionality someday, just like this!  Thanks for putting that together here. :)

Hoikas:
This video seems to be distorted; the original from TPotS is 640x480, but this one is 1280x720, resulting in noticeable artifacting.

Hazado:
we can increase the size of videos via python

ZarothYe:
> This video seems to be distorted; the original from TPotS is 640x480, but this one is 1280x720, resulting in noticeable artifacting.

I thought I was using the same size and dimensions that Haz's video from the Cleft Start work provided -- I used that file as the base file to mix the additional audio tracks into, but I'm sure it's possible I did it wrong. The dimensions can be adjusted with the tools I was using. Just tell me what you want it to be. Also... there was always noticeable artifacting, I think, even in the original .bik file. I did a visual comparison of this one, Haz's, and the .bik from my CC folder, and didn't notice anything different. Again, I could be wrong.

ZarothYe:
Talked to Adam briefly about this. I have uploaded a new version generated directly from the original asset video and using the same dimensions and framerate as the original. Please take a look and see if the artifacting issue is better, because I don't have a good eye for it. As a possible bonus, the file is now 16MB -- about half the size of my first version. I chose a middle-of-the-road quality value for generating the output, but I can choose something higher quality if needed. The source and this result looked the same to me, but as I said, I don't have a good eye for it.

Hoikas (approved):

PR #166, opened 2022-07-05
Add 6 new fonts (5 handwriting/script + 1 print) in even sizes 10 to 30

ZarothYe:
Adding new options for journaling fonts. All are developed by the Uyghur Computer Science Association (UKIJ) and released under GPL/LGPL. Looks like they made good quality fonts with good support for Western European characters, although some have missing support for a few accented characters and punctuation like ä, ¿, and ¡.

UKIJ Jelliy: https://fontlibrary.org/en/font/ukij-jelliy
ukijje_sample18

UKIJ Moy Qelem: https://fontlibrary.org/en/font/ukij-moy-qelem
ukijmoyq_sample18

UKIJ Qolyazma: https://fontlibrary.org/en/font/ukij-qolyazma
ukijqolyazma_sample12

UKIJ Qolyazma Tez:
ukijqol_tez_sample16

UKIJ Qolyazma Tuz: 
ukijqol_tuz_sample16

UKIJ Qolyazma Yantu:
ukijqol_yantu_sample16


Hoikas (commented):
I'm thinking these fonts probably need to be renamed to have slightly more artist-friendly names. There is precendent for renaming the p2f version of a font in that fonts like "Michelle" and "Nick" can be found using services like [whatthefont](https://www.myfonts.com/pages/whatthefont) under different names. Also, I think it would be a good idea to place the TTFs into a subdirectory with an extra file pointing to the source the fonts came from.

ZarothYe:
I already provided the TTFs in the sources/fonts subdirectory. Are you saying you'd like a text files in that subdirectory with a sort of reference for all of them?

I'm also not good at naming things and kind of felt like I might be a bit of a jerk by unilaterally deciding to obscure the creator's original name. So, I kept the TTFs as the original names as downloaded from the source, and the TTF name is the default choice for the P2F files when going through `plFontConverter`. I figured it was one less step to remember to do if someone wanted to generate more sizes of the fonts.

ZarothYe:
Renamed the p2f files and added a subfolder with a reference text file for the UKIJ TTFs, as requested.

Hoikas (approved):

Hoikas:
No objections in over a month. Merging.

PR #172, opened 2022-08-03
Garrison Updates to Match Moula Q2

Hazado:
Removed panel red lights fade out and back in when they are used, makes the blocker counters look a little better.
Made blocker screens not flicker anymore and half as transparant.
Flashing lights when book deploys now flash half as slow but the entire time the book is out
Made Building 2 map screen in conference rooms not flicker anymore and half as transparant.
Added visregions to all wall viewer screens in observation rooms to help with age lag
Reduced static opacity for observer rooms and nexi
Fixed the Green Button animations for both panels
Fixed bug with when the green buttons were supposed to be able to be clicked


Hoikas (approved):
In the future, we would prefer to see these as atomic changes grouped by topic.

PR #173, opened 2022-08-07
Add highgarden

DoobesURU:
Adds the PRPs and LOC file for Tweek's Age of Fahets: Highgarden.


Hoikas (approved):

ZarothYe:
It doesn't look like this included my suggested edits/corrections for typos and such for the LOC, possibly. It seemed like Tweek might have wanted to include those but submissions had already been frozen for OU so it didn't make it into MOULa yet. Would it make sense to include that here since there is no freeze deadline, or indeed, no requirement that the LOCs match those in OU at all?

Hoikas:
It would be nice to have those, I think.

DoobesURU:
> It doesn't look like this included my suggested edits/corrections for typos and such for the LOC, possibly. It seemed like Tweek might have wanted to include those but submissions had already been frozen for OU so it didn't make it into MOULa yet. Would it make sense to include that here since there is no freeze deadline, or indeed, no requirement that the LOCs match those in OU at all?

I found your changes on the Discord thread. Give me a bit of time and I'll incorporate them for a new commit.

Hoikas (approved):

DoobesURU:
Shoot. It looks like I may have broken some of the CDATA bracketing on one of the elements. Please hold off on merging until I can get home later today and fix it, please!

Hoikas:
No objections noted in a little over a week. Merging.

PR #175, opened 2022-08-20
Adjust Hoods' Podium Python Mods

DoobesURU:
Sets the Python file mod for the podium to proposed xPodium.py script (see https://github.com/H-uru/Plasma/pull/1238 ).

This is mostly for a test given @Hazado will likely have further updates for next quarter's Kirel audio.


DoobesURU:
For reference, here's the revised Python file mod in PRPShop: 
image

Hoikas (approved):

PR #184, opened 2022-11-22
Fix shading on the hydrant Cleft book in Relto

dpogue:
Still not sure why the shading is wonky on this particular book since it matches how all the other books are set up, but it's particularly bad. The easy fix is the adjust the ambient (emissive) colour of the layers slightly to make it appear brighter.

Result in-game:
Screenshot from 2022-11-21 20-12-23


ZarothYe:
I've noticed this before and it always sort of bugged me. This fix is exciting! 😄

Hoikas (approved):

Issue #186, opened 2022-12-06
Zandi's Radio Playback Issue


patmauro:
When **clftZandiRadio_Loop.ogg** was replaced (presumably to avoid a licensing issue) the replacement file introduced a few issues affecting sound playback in-game:
1. The audio plays back at half the speed it should - The original file was mono, wheras the replacement was stereo. Because this is 2x the information to stream, it means the track plays back at half the intended speed. (I think nobody noticed this because slowed-down violins sound like cellos! But this is, in fact, wrong!)
2. The audio was not properly filtered to correctly mimic the sound of a tiny radio. At the moment, Zandi appears to have a hi-fi system hidden somewhere in his airstream.

While the PRPs could be adjusted to account for this, in this case it is actually better to leave these alone and simply make the corrections file-side - this sound _should_ be mono; this is actually correct.

**I've prepared a corrected OGG file - contained in the attached ZIP - which addresses both issues**; just need a pull request for it, etc. Here is a demonstration of the fix captured in-game: https://youtu.be/1caVODpUcEk

Note that while there might be value in introducing a legally-distinct "soundalike" of BYUBYD in the future (which might be more appropriate for Zandi), this would take a considerable amount of time & effort to put together - so **there is still value in solving the underlying issue here first before moving on to any enhancements**.
[clftZandiRadio_Loop.zip](https://github.com/H-uru/moul-assets/files/10167361/clftZandiRadio_Loop.zip)

DoobesURU:
I think I encountered a similar situation with the Relto cannen when getting it ready to play Veelay music. Glad you caught that!

PR #187, opened 2022-12-07
Cleft zandi radio fix

patmauro:
Fixes https://github.com/H-uru/moul-assets/issues/186 by:

- Archiving existing file
- Replacing OGG with a new OGG mixed in mono for proper playback and with an appropriate filtering effect.


Hoikas (commented):
Please delete `compiled/sfx/clftZandiRadio_Raw_Stereo.ogg` - it can be retrieved from git history if needed. Further, `compiled/sfx/clftZandiRadio_Loop.ogg` seems to be a binary file in the git repository instead of properly stored in git LFS. Please ensure that this file is stored in git LFS due to repository size limitations.

Hoikas:
I've gone ahead and corrected the repository management issues cited above.

Hoikas (approved):

PR #189, opened 2022-12-08
Chiso Preniv Q3-4 2022 Updates

DoobesURU:
* Brings Chiso up to current MOULa version


Hoikas:
These two linking books show a black panel in world and do not have hotspots.

image

Hazado:
Pretty sure those are blanks for new books.
@DoobesURU moved two books from the center ring to the second floor permanently.

DoobesURU:
> These two linking books show a black panel in world and do not have hotspots.
> 
> image
> 

Yeah, my bad. I neglected to mention that the Pub and Veelay Books are now upstairs, so those two spaces are meant to be off until we get new Ages to fill them.

Hoikas (approved):

PR #190, opened 2022-12-16
Adjust *.loc files to have LF line endings instead of CRLF

ZarothYe:
After pulling fresh bits from master, I got the dreaded notification from `git status` that my copy of VeeTsah.loc was modified automatically during checkout. This, in my experience, typically means there was an inconsistency in the line endings from what was committed (looked like CRLF line endings) and what was intended to be in the repository (LF line endings). I opened the LOC in Notepad++, did an EOL conversion to LF and re-saved. I touched nothing else. Git's comparison is eminently unhelpful, unfortunately.

If there's a better way to solve this, please let me know. :)


Hoikas:
> If there's a better way to solve this, please let me know. :)

Don't merge the wrong line endings 😭 @Hoikas

ZarothYe:
Based on results from Hoikas's new toy (https://github.com/H-uru/moul-assets/pull/193), I have pushed Serene.loc through plLocalizationEditor to turn it appropriately into UTF-8 (also CDATA as a bonus side-effect). Changed Tiam.loc line endings from CRLF to LF as well. PR to update Serene.loc to appropriate latest version coming separately.

Hoikas (approved):

Hoikas (approved):

Hoikas:
✔️ Yay for green checks 🎉

Issue #188, opened 2022-12-07
Phil's Relto - rain effect starts late


patmauro:
This has been discussed previously, but just to officially log this - in Phil's Relto, the rain effect begins late, about 30 seconds after linking into the age rather than right away.

It is possible something may have been incorrectly ported here from Rainy Cleft, as this 30ish-second interval appears to match the same 30ish-second pause that occurs before the rain effect begins in that age for dramatic effect. However in that case the pause is clearly intentional, and there are matching changes in ambient sounds, water ripple effects, etc to time with this, whereas in Phil's Relto this appears to be a clear mistake - we hear rain, but do not see it.

dpogue:
Does this happen with the rainy page in (normal) Relto too?

patmauro:
> Does this happen with the rainy page in (normal) Relto too?

No; behaves correctly in normal!Relto as far as I can tell.

Hazado:
The hsBezScalarKey in the "RainParty_rain_anim_0" ATC Anim start with a value of 0 which means no particles.
Just need to change the value to 600.
I can submit a fix for this soon, if no one else gets around to it.

PR #192, opened 2022-12-16
Phil Relto rain starts immediately

Hazado:
Fixes the Phil Relto rain so it starts immediately instead of after 30 seconds.

Diff:
[PhilRelto.txt](https://github.com/H-uru/moul-assets/files/10249537/PhilRelto.txt)


patmauro:
Note - this fix should resolve https://github.com/H-uru/moul-assets/issues/188

Hoikas (approved):

PR #200, opened 2022-12-31
Textures added and changed to allow age reset mechanic from books

Hazado:
Bookmark textures provided by by Keith "Tweek" Lord.

image
image


Hoikas (commented):
I like this addition! Could you also commit the original texture images from Tweek as well?

Hazado:
> I like this addition! Could you also commit the original texture images from Tweek as well?

Original textures are submitted.

Deledrius:
I'm not sure an IC mechanism for an OOC function makes a lot of sense in this way, but I'm at a loss for a better way to do this and there's certainly a need for it.  

It just doesn't make sense to me that a bookmark would "reset" an Age in any way, especially outside of the very specific Yeesha-interfaces we've been given.  On the other hand, it's definitely not the only concession made to gameplay like this in the history of the game.

Is the top image's bookmark supposed to match the second image?  Not only are they slightly different in texture, but the top one appears to have a slightly-wrong aspect ratio.

dpogue:
> Is the top image's bookmark supposed to match the second image?

No, there are two different choices for bookmark style available for Age authors to pick.

Deledrius:
Ah, okay.  I thought that might be the case but wasn't sure if that was the intent or not.

dpogue:
> It just doesn't make sense to me that a bookmark would "reset" an Age in any way, especially outside of the very specific Yeesha-interfaces we've been given. On the other hand, it's definitely not the only concession made to gameplay like this in the history of the game.

Yeah, this is mostly a gameplay concession. Right now resetting happens via the 0 button on the Relto bookshelf, but fan Ages don't go onto the Relto bookshelf so there's currently no way for a player to reset their instance of Ages like Tiam and Elonin.

The bookmark reset idea is an attempt to solve that problem, while keeping the reset mechanism attached to the book (in Chiso Preniv)

Hazado:
These are the Tweeks original bookmarks. The current fan ages have all chosen to use the second of the textures.
image
image

DamnBriggsy:
I believe that this is an elegant solution to a problem that definitely needed one. I'm all in favor of the submitted designs and functionality.

Hazado:
> Not only are they slightly different in texture, but the top one appears to have a slightly-wrong aspect ratio.

I did try very hard to get the aspect ratio close to looking like the originals, but due to how the books are assembled from jpegs using book html... thats a hard thing to do. I think it looks pretty close to the originals

Deledrius:
I really like the designs visually.  It's interesting that the images aren't stretched here, but the one is in the sample image in-situ.

Like I said, I'm unhappy with the mechanism eroding the IC barrier here, but it's clearly needed.  Perhaps with some brainstorming there could have been some other method, but this works as-is and looks nice doing it.

Deledrius:
> The "bahro stone/yeesha" style one could probably be made a little wider

Oh!  Now that I see it up close it's easier to read that this is a Yeesha/bahro weave.  I should have noticed that.  This definitely closes the IC concern, or at least greatly narrows it.  Nice work by Tweek there.

Hazado:
Okay pushed a fix to the first bookmark to fix the aspect ratio
image

Deledrius:
Looks great to me!

Hoikas (approved):

PR #201, opened 2023-01-08
Switches SwimDockExit and WallClimb* to Global Animations

Hazado:
These animations are swapped to global animations so that they can be used in multiple ages.
Remove SwimDockExit from psnlFanYeeshaPages01.prp as they would cause issues in Ahnonay

Procedure used to swap them
1. Determine new ID for the pages
2. Load "GlobalAnimations.age" in PRPshop
3. Open specific animations as well in the same PRPshop
4. Show Properties
5. Change the ID to -1, and the new page ID chosen in step 1
6. Rename the Scene Node in the page to GlobalAnimations_*
7. Save the file in PRPshop
8. Rename the file in the directory to GlobalAnimations_*.prp


Hazado:
Included a version of the original animation files with all the ATC stuff removed. They should all be 1kb files now.

Hoikas (approved):

PR #204, opened 2023-01-10
Restore DRC Stage functionality

DamnBriggsy:
Restores the option to change the state of the stage in the courtyard via SDL Thanks in large part to Hazado for showing me which python file controls the stage and leading me to what needed to be added to the prp to make it work.

city.sdl's `islmDRCStageState` can be set to 0 (off), 1 (on), or 2 (on with DRC decorations)


Hazado:
Has a lot of changes to other objects in this file.
Don't think this PRP was based off of master and came from somewhere else

DamnBriggsy:
state 0 

state 1

state 2

DamnBriggsy:
> Has a lot of changes to other objects in this file. Don't think this PRP was based off of master and came from somewhere else

Should be in good order now, thanks again!

Hazado (approved):

DamnBriggsy:
Per Hazado, `<plKey Name="city_ferry" Type="plSceneNode" Location="6;0" LocFlag="0x0000" ObjID="1" />` was added to the SceneNode modifier of the SceneObject `SDLAgeHook`.

DamnBriggsy:
@Hazado changed the collider in this spot to allow for people to walk back and use the stage. Thank you!
collider

Hazado (approved):

DamnBriggsy:
Additional changes by @Hazado following a round of testing:

- Rope fence colliders lowered on State 1
- Step up collision improved
- Collision added to boxes
- State 2 rope fence colliders reverted to original

Justifications for changes:
The camera behaved poorly when interfacing with the collision on the rope fence
H'uru client would get stuck on step up
Mild discomfort caused by intangible boxes
Creates the possibility of an "official" use of the stage by making it generally inaccessible for normal users while the DRC decorations are enabled.

DamnBriggsy:
New version by @Hazado, box colliders adjusted slightly smaller.

Hoikas (commented):
The new scene object with the emgrphase0 python file mod needs to have its synch flags set correctly:

```xml
	<SyncParams flags="0x00000010">
		<ExcludePersistentStates>AGMaster Responder Layer Sound XRegion </ExcludePersistentStates>
		<VolatileStates></VolatileStates>
	</SyncParams>
```

The current setup will cause extra SDL states to be sent to the server.

DamnBriggsy:
> The new scene object with the emgrphase0 python file mod needs to have its synch flags set correctly:

An attempt was made to resolve this, but I may have made a mess of this PR in the process.

Hoikas (approved):

PR #206, opened 2023-01-19
Add Video Selector buttons for Video Options Menu GUI

Hazado:
This is preparation for the when we have more than 2 in-game videos.

Changes the "Trailer GUI" to have a previous and next button
Starts playing the first video in the AVI directory
Changes the Options Menu "Uru Live Intro" to use localization text
Change the Localization text to "Videos"

Trailer GUI:
image

800x600:
image

Options GUI
image


Deledrius:
Will this only affect the video viewer when launched from the Options menu, or is it visible on the first-play intro screen as well?

Hazado:
> Will this only affect the video viewer when launched from the Options menu, or is it visible on the first-play intro screen as well?

It does at the moment, need to fix that...

Deledrius:
Since they have a black background, it would be nice to have a bit more padding around the `previous` and `explore` buttons, and a little less on the `next` button so that they appear more uniform.

Hazado:
> Since they have a black background, it would be nice to have a bit more padding around the `previous` and `explore` buttons, and a little less on the `next` button so that they appear more uniform.

The buttons are actually all the same size

Deledrius:
I see that.  It leaves the `next` button with unfortunate empty space, both before and after the word, whereas the others do not.

Hazado:
Okay, the character intro video does not show any of the new buttons now

Hazado:
> I see that. It leaves the `next` button with unfortunate empty space, both before and after the word, whereas the others do not.

Would you prefer that `next` was touching the left instead and had the black space on the right?

Deledrius:
Transparent looks much better, I think.

Deledrius (approved):

Hazado:
Had to resave all the GUI files due to loc keyring issues found when using an external client.
The options menu GUI apparently has a lot of interdepencies for other GUI files

PR #208, opened 2023-01-25
Add Cannen Filter to Veelay Music

patmauro:
With Mystler's blessing, retroactively applies the new cannen filtering effect to Relto's Veelay track. Going forward, these will be standardized and all new cannen tracks should use this same filter.

I apologize I was unable to properly upload this with Git LFS - I ran into issues getting this to work and had to resort to GitHub frontend instead. I will need someone to make the needed correction for me.

### DIFF
--- a/compiled/sfx/psnlVeeTsahMusic.ogg
+++ b/compiled/sfx/psnlVeeTsahMusic.ogg
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:98ca2577bfbf812e94c74a5f606d18f4d8417174b6bc9e0f871f40a66c2e1967
-size 3817928
+oid sha256:0eef431586ad44bda251e93099821e72216d0ea2cfab7486b4d45a7ff4f363e5
+size 4279212


patmauro:
Per Doobes, needs to be updated with a smaller filesize bounce - I'll take another stab at pushing the change via Git LFS.

Hoikas:
Just to verify - you ran the original uncompressed .wav (or .flac) through the filter to produce the new .ogg?

patmauro:
Normally this _would_ be the standard, but no, in this case we just used the existing OGG - I don't think we need to be too concerned about loss (or compounded loss) in this particular case given the filter we're applying is already limiting the dynamic range pretty significantly (on purpose). As is, we had to reduce the export quality quite a bit in order to find a happy medium that produced reasonable file sizes.

We can circle back and try to fetch an original lossless bounce from Mystler, but I think that would probably be overkill and any benefit negligible - in this instance, I think what we have is sufficient.

Hoikas:
I would prefer to have the file generated from the original source before merging this.

patmauro:
> I would prefer to have the file generated from the original source before merging this.

Understood; let me see what I can do.

Hoikas:
This PR breaks the TL deployment:
```
error: Your local changes to the following files would be overwritten by merge:
        compiled/sfx/psnlVeeTsahMusic.ogg
Please commit your changes or stash them before you merge.
Aborting
fatal: cherry-pick failed
Encountered 1 file(s) that should have been pointers, but weren't:
        compiled/sfx/psnlVeeTsahMusic.ogg
warning: You seem to have moved HEAD. Not rewinding, check your HEAD!
```
Please remove the first commit with the in-repo sound so that we can proceed.

Hoikas (approved):

PR #209, opened 2023-01-25
Wall Game fixes

Hazado:
Add camera responder to logic modifiers for chair clickables 
Adjust Region for panel clickables to be much larger

Diff File - [WallPhy.txt](https://github.com/H-uru/moul-assets/files/10496627/WallPhy.txt)

### DIFF
--- a/compiled/dat/Garrison_District_WallAdditions.prp
+++ b/compiled/dat/Garrison_District_WallAdditions.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6fbc38da2052f468251392476cb04dbe1f03e28f0f81e7e1589d6fe5e281e7cb
+oid sha256:7f40a35ac177213359004e79db5f3b55071b0ba51ee7f217dc42be616e75cb69
 size 83873
--- a/compiled/dat/Garrison_District_WallAdditions.prp
+++ b/compiled/dat/Garrison_District_WallAdditions.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:1d3af6f47055a80545f75b9f7d93006d781d4f723b1a78cbb77b8900922fcffb
-size 62128
+oid sha256:6fbc38da2052f468251392476cb04dbe1f03e28f0f81e7e1589d6fe5e281e7cb
+size 83873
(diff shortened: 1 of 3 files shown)


Hazado:
Changed the conference room camera views to a load mask setup.
Ultra Graphics Quality setting shows 1024x1024 at 0.1 refresh rate
High Graphics Quality setting shows 256x256 at 0.05 refresh rate
Medium Graphics Quality setting and lower wont show the camera views at all.

Removed these light flares from the camera views inside the conference room
image

Diff Files
[walladditionsdiff.txt](https://github.com/H-uru/moul-assets/files/10609204/walladditionsdiff.txt)
[wallroomdiff.txt](https://github.com/H-uru/moul-assets/files/10609205/wallroomdiff.txt)

Hoikas:
FWIW, your high setup will render more frequently than the ultra one. I would expect for the ultra quality image to render at basically 60fps (about 0.01), and for the high one to render less frequently eg maybe 30fps (about 0.03). You don't have to use those exact numbers, though. I expect both of those resolutions could safely be halved as well...

dpogue:
> I expect both of those resolutions could safely be halved as well...

Just noting (because it came up previously when Hazado and I were talking about DCM resolutions): on H'uru clients, DCMs will always be rendered at the full resolution of the pipeline, regardless of the resolution provided, if pixel shaders v3 is supported.

Hoikas:
Good point. We may want to add an escape hatch for that in the case of DCMs that aren't reflections... or have some other way of specifying it.

Hazado:
> FWIW, your high setup will render more frequently than the ultra one. I would expect for the ultra quality image to render at basically 60fps (about 0.01), and for the high one to render less frequently eg maybe 30fps (about 0.03). You don't have to use those exact numbers, though. I expect both of those resolutions could safely be halved as well...

my intent was to reduce the load of the rendering, The wall climbing isnt exactly fast moving.

The resolutions are more for the OU client because Huru doesnt use them.

If you think the new refresh speeds wont cause issues for clients, Im fine with upping it to 60fps amd 30fps

dpogue:
> Ultra Graphics Quality setting shows 1024x1024 at 0.1 refresh rate
> High Graphics Quality setting shows 256x256 at 0.05 refresh rate

I think what Hoikas was implying is that smaller numbers mean a higher refresh rate (shorter time between frames), so High at 0.05 is actually going to update more often that Ultra at 0.1, which is probably not what you intended?

Hazado:
I've made the change so the Ultra runs at 60fps and High at 30fps.

Hoikas (approved):

PR #213, opened 2023-02-03
Fixes KI Light Sprite visibility in first person

Hazado:
Added a Distance Fade modifier to the HandSprite for the KI Light.
This will cause the KI Hand Sprite to not be visible if the camera is within 3 units of it.
Effectively, it hides it while in first person and makes it visible in third person.

The current KI light scripts do not netforce, so it is very hard to see the hand sprite on someone else in normal gameplay.
If you did happen to see one, you need to be very close to the other avatar for it to disappear. Basically, standing right on top of them and looking down at their hand before it will vanish.

Before:
image

After:
image

Only change made was to add the Distance Fade Modifier in 3dsMax. 
(Changes the icicle min distance in the drawable span from 0 to 3)
The Max file covered both the GlobalAvatars_District_[Fe]Male.prp and the GlobalClothing_District_[Fe]Male.prp. The Fan02 prps were resaved due to having loc key changes in the GlobalClothing M|F file from the export from Max.


Hoikas:
Do the changes still work if you omit the export of the GlobalClothing files?

Hazado:
globalclothing prps would have keyring issues I think.

Hoikas:
They shouldn't; the GlobalAvatar files do not reference GlobalClothing AFAIK.

Hazado:
Ill double check that the globalclothing files dont reference anything in globalavatars tomorrow and remove if they dont

Hazado:
Yeah, the plSharedMesh uses plMorphDataSet from GlobalAvatars inside GlobalClothing_District_[Fe]Male.prp

Hoikas (approved):

Issue #222, opened 2023-02-20
Corrupted Files?


MaxwellEdison:
Hey, I've tried to download the project, however it appears as though all of the file content has been replaced with text.  For an example, attempting to open the AhnonayCathedral.blend in blender results in an unsupported file error.  If I open it in notepad++, it gives me:

version https://git-lfs.github.com/spec/v1
oid sha256:fcdbfab5f89942c899eaf8b7f1eb01a3a2ab57bec6f06daab413d5d8c553d451
size 23180444

All of the other files appear to be similar. 

Is there something I need to do in order to be able to download the actual files?  Are these files hosted elsewhere?

bertvandepoel:
Hey Maxwell. For moul-assets, the H'uru project is using LFS (Git Large File System) to store the actual files, as they are too large for GitHub. I suggest you use "git lfs clone" instead of "git clone". That ought to help. Let us know if that fixes things! :)

MaxwellEdison:
Alright, it looks like it's downloading the files now.  If I may make a suggestion, you might want to mention that you're using Git LFS and a quick "download git lfs here and run this command" tutorial in the readme for people who aren't familiar with github/git lfs.

Deledrius:
@MaxwellEdison Good suggestion.  We have instructions for contributing but not the initial clone.  Whoops!  Thanks for pointing that out.

PR #223, opened 2023-02-21
Add instructions for browsing using LFS cloning

bertvandepoel:
Based on #222 there was a need for this. Not sure if the "git lfs clone" is the best option, or if it would be better to just go for a "one time git lfs install", so feedback is very welcome!


README.md
```diff
@@ -7,6 +7,9 @@ This repository is used to host assets used in Cyan World's Myst Online: Uru Liv
 * [dirtsand](https://github.com/H-uru/dirtsand) - An open-source Plasma-compatible server project.
 * [korman](https://github.com/H-uru/korman) - Blender plugin for creating ages for Cyan Worlds' Plasma engine.
 
+## Browsing MOUL Assets
+You may notice while browsing this repository online or after cloning on your own machine that most files seem to be small text files containing some kind of hash reference. This is because the MOUL assets repository is too large for (free) GitHub. We therefore use an external Git Large File Storage server. The easiest way to download the entire repository is to use `git lfs clone`. Do keep in mind that the repository is very large (over 20GB), so make sure to run this command in a location with sufficient space. You can of course also enable Git LFS be default. You can find some basic information on the [Git LFS website](https://git-lfs.com/) or refer to the [Git LFS wiki page on pulling and cloning](https://github.com/git-lfs/git-lfs/wiki/Tutorial#pulling-and-cloning).
```
dpogue, line 11: > You can of course also enable Git LFS be default.

This should say "**by** default", and we might also want to drop the "of course"

Deledrius, line 11:
```suggestion
You may notice while browsing this repository online or after cloning on your own machine that most files seem to be small text files containing some kind of hash reference. This is because the MOUL assets repository is too large for (free) GitHub. We therefore use an external Git Large File Storage server. The easiest way to download the entire repository is to use `git lfs clone`. Do keep in mind that the repository is very large (over 20GB), so make sure to run this command in a location with sufficient space. You may also choose to enable Git LFS by default. You can find some basic information on the [Git LFS website](https://git-lfs.com/) or refer to the [Git LFS wiki page on pulling and cloning](https://github.com/git-lfs/git-lfs/wiki/Tutorial#pulling-and-cloning).
```

Deledrius (changes requested):

Deledrius:
According to the docs (and in my own testing), it appears that the `git lfs clone` command is unnecessary once git-lfs is installed.  On Windows, git-lfs is installed by default with git-for-windows, but generally needs to be installed as its own package on Linux and macOS (instructions are provided in the linked documentation).  Once that's done, a normal `git clone` appears to work as expected against a repository with LFS data.

Hoikas (approved):

PR #225, opened 2023-03-05
Fix Incorrect Textures In (Old) Messenger's Pub

patmauro:
Resolves Issue https://github.com/H-uru/moul-assets/issues/168 by making the following corrections:

1. Replaces the GoG logo in gpmstapestrysymbol_0#0.hsm with the Guild of Messengers logo; also adds a background texture to make this easier to modify rather than rely on alpha channel. (Thanks to @DoobesURU for the help solving some of the odd texture issues here.)
2. Flips gpmscolumnmessengerspaint180_0#0.hsm horizontally, so the linking panel shows up in the correct spot on the 2nd floor wall texture.

This specifically resolves the issues brought up in https://github.com/H-uru/moul-assets/issues/168 and is separate from the more general Pub revamps @DoobesURU will be pushing in the future.

GoMePubFixed_small


patmauro:
Texture quality review: 
- Wall texture is good.
- Banner texture is slightly degraded compared to original; this largely will not be an issue since this is more of a background detail players won't see up close, but it may be worth seeing if it is possible to get better fidelity via a different compression method, as with PR227.

patmauro:
Optimized the banner texture with @Hoikas' high-fidelity compressor and added the new texture source to assets. We should now be ready to close this PR.

Hoikas (approved):

PR #227, opened 2023-03-09
Supplemental Update to PR224 (Phil's Relto Bookshelf)

patmauro:
A follow-up to PR https://github.com/H-uru/moul-assets/pull/224, as we received some testing feedback after this PR was already merged & closed.

This second commit addresses the following:
1. Adjusts alignment of Kirel book to match legacy as closely as possible from the new position on the texture sheet.
2. Fixes a gap in Cathedral book's spine.
3. Switches to an uncompressed DDS for the book textures - after multiple tests, I confirmed that _any_ level of compression seems to introduce noticeable image degradation to the book spines, which is particularly unfortunate in this case as these assets will be directly in front of the player camera. This seems to be unavoidable due to the fact that we do not have the original uncompressed texture sheet to work from. Switching to an uncompressed dds will increase the filesize of the prp a bit, but in this case I think it is worth it.
4. Note - I reviewed the alignment of all other book textures and confirmed that, while some may be a little more squished compared to standard Relto due to mesh differences, the positioning of all book textures should indeed be optimal as is.
5. Note - These changes did _not_ appear to require an additional update to **philRelto_District_PhilsRelto** in my tests, so a new version of this file is not included in this revision. **However, please review and confirm.**

Recommend keeping this PR open for another week or so, just in case there's any additional testing feedback (but I don't anticipate any at this point). As long as there are no additional notes by 3/15, I'm good to close.


Hoikas:
> Switches to an uncompressed DDS for the book textures - after multiple tests, I confirmed that any level of compression seems to introduce noticeable image degradation to the book spines, which is particularly unfortunate in this case as these assets will be directly in front of the player camera. This seems to be unavoidable due to the fact that we do not have the original uncompressed texture sheet to work from. Switching to an uncompressed dds will increase the filesize of the prp a bit, but in this case I think it is worth it.

We strongly prefer to use compressed mipmaps to prevent texture bloat, especially in Cyan Ages. The compression artifacting can probably be fixed by manually using the high quality compressor available in libHSPlasma, as was done for the Gay Zero shirt.

patmauro:
I'm certainly willing to give the high quality compressor in libHSPlasma a shot. If I'm happy with the results, I may even update PR 225 to take advantage of this as well (although this is less important in that particular case).

patmauro:
Reworked the texture sheet to include the lossless Intangibles Assets where applicable so that compounded image degradation is less of a concern. Thanks to an assist from @Hoikas we have also re-exported with the high quality compressor as well. I've have also added the uncompressed texture to our texture assets collection for reference.

@Hoikas - we should be good to close now barring any additional comments or issues. OU will be sent the equivalent update in Q2 (they currently are synced with commit [e7dfc0f](https://github.com/H-uru/moul-assets/pull/227/commits/e7dfc0f63a25006e57efc0119871fc0a818f3785))

Hoikas (approved):

PR #238, opened 2023-06-15
Asset Blend Files

DoobesURU:
This is an initial commit of blend files containing assets from various Cyan Ages (and also the new GoMe Pub).

These include:

* Zandi's chairs from the cleft (two files: single chair and two chairs)
* Several types of construction cones and barricades
* The "crazy lamp" found in Ae'gura in places such as outside the museum door
* The "flat" lamps from the hoods (Bevin, Seret/Kirel, and a blank one)
* Three different gazebo files (roof w/ lights on, roof w/ lights off, no roof w/ Linking Book)
* Various banners from the new GoMe Pub (Guild, Cyan-related, Pride, etc, plus one blank)
* Guild tapestries from the five Cyan Pubs (plus one blank)
* The entryway from the hoods
* Lamp found in a few of the Eders
* Two versions of the Nexus pedestal: one "hooded" (like next to library in Ae'gura) and one for walls (ie no back mesh)
* Simple Book pedestal (eg the pedestal next to the KI machine in Gahreesen with the Nexus Book)
* Signpost with arrows from Ae'gura
* Large ornamental statue from various garden Ages.

The blends come with both the license and instructions displayed and should contain all the needed textures. If something is missing, that can be remedied in future commits.

Also, I took a guess as to what we'd want for the file structure, but that can be easily modified too, of course.

Any thoughts and suggestions on this new process are most welcome!

### DIFF
--- a/sources/blends/CleftChair.blend
+++ b/sources/blends/CleftChair.blend
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:39575e83261a2f8cc0c96c10017b9c95b44bf47fc03e4428a802917278d94d9e
+size 3675672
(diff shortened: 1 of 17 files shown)


Deledrius:
Nice work putting these together!

PR #239, opened 2023-07-22
Remove references to EAX from player-facing UI

patmauro:
Admittedly, this may be jumping the gun a little bit, as https://github.com/H-uru/Plasma/pull/1000 is not complete yet - however, given that EAX is *very* much dead and buried, and much of Uru's sound logic has _already_ been ripped out in favor of OpenAL, it still feels like it's as good a time as any to go ahead and remove the EAX references from the player-facing UI once and for all. This PR accomplishes that by (A) removing the Creative EAX logo seen at the bottom of the Audio Settings menu in favor of an OpenAL graphic, and (B) and changing some of the Audio Settings menu's text.

image

Some notes:
- The original plan was to put this in a brown-toned box like the EAX logo, but this didn't quite look right after seeing it in context. Instead, I opted to *invert* this so it fees a bit more like a natural part of the interface.
- Flat color felt wrong here too, so I reintroduced the two-tone color usage in OpenAL's logo but was very careful to keep this in a color scheme matching the UI.
- URU slightly stretches out the logo texture in-game. To compensate, I squeezed the logo in the texture sheet to make sure it displays at the correct size.
- With the **"Enable EAX"** entry, I'd probably have _preferred_ to simply hide it outright (temporarily, of course, until such time as a proper EFX enhancement is available), but I wasn't able to figure out a clean way of doing this I felt comfortable integrating. So instead, I have changed the text so it at least reads "EFX" instead - this will at least save us a step later on, and the option is still disabled either way. This should apply to all languages.
- I've verified there doesn't appear to be any generational loss to the texture sheet after swapping out the graphic, so we should be OK. However...
- I'd still rather source from a lossless version of this texture sheet, but unfortunately, as far as I can tell, we don't have one we can source from - the version of this in our lossless source files is the iteration from CC, which is significantly different. That said, if anyone is aware of a lossless version of the MOUL iteration, please let me know.


Hoikas:
I think I would prefer to not have any sound related logo at all. Further, we prefer to see changes to the KI (the options menu is part of the KI in terms of max files) be done in xKIGUI.max and reexported afresh so that our source files can reproduce the game files exactly.

patmauro:
> I think I would prefer to not have any sound related logo at all. Further, we prefer to see changes to the KI (the options menu is part of the KI in terms of max files) be done in xKIGUI.max and reexported afresh so that our source files can reproduce the game files exactly.

I guess that makes sense. My idea here was that we could just handle this with a 1-1 replacement of existing assets rather than a rebuild, but if there's a cleaner way to do this, then that's probably the way to go. And I do agree: as happy as I am with how the OpenAL replacement graphic came out, no logo at _all_ would, frankly, probably be best.

Editing and re-exporting max files is a little outside my wheelhouse though, so someone else would need to take point there if we wanna go that way.

patmauro:
As this has been more or less rejected, I am closing this PR and have opened Issue https://github.com/H-uru/moul-assets/issues/247 to cover this topic instead.

Hoikas:
> As this has been more or less rejected

I think it's more that we are waiting for someone with Max to go in and reproduce this change. That person is probably me, unfortunately, and opening Max is something of a chore 😞

patmauro:
> > As this has been more or less rejected
> 
> I think it's more that we are waiting for someone with Max to go in and reproduce this change. That person is probably me, unfortunately, and opening Max is something of a chore 😞 
> 

No worries, I didn't take it personally ;) but a different approach to this fix will require an altogether different PR, ultimately, so I figured for the sake of clutter there was no need to keep this one open - in the meantime we can treat as an Issue.

Issue #33, opened 2021-04-27
Fix the city doors (library & museum)


dpogue:
These have always been a synchronization problem, but I seem to recall a certain mookow getting them working properly for UU.

Would be nice to reinstate those doors and have them working as intended.

Deledrius:
I would love to see these finally fixed!

patmauro:
This has been resolved by https://github.com/H-uru/moul-assets/pull/242 and https://github.com/H-uru/Plasma/pull/1462 and can now be closed! 😃

PR #242, opened 2023-08-13
Adds back the doors to the City

Hazado:
Script files - https://github.com/H-uru/Plasma/pull/1462

Adds back the doors to the city age
Library - https://www.youtube.com/watch?v=eQN9g3_GvGw
Museum - https://youtu.be/cKd7t_hcM-A

Added actual collisions to the doors
Moved the exclusion region on/off to the beginning of the responder

These changes make it so your not running into invisible barriers when the door is opening. When the door is closing you will notice invisible barrier a little bit, but without that I was able to be pushed through the wall and fall through the ground to the basement. I feel that these changes make the doors a better experience than what happens in CC.

### DIFF
--- a/compiled/dat/city_District_LibraryInterior.prp
+++ b/compiled/dat/city_District_LibraryInterior.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:2a5288e76cac0a5a70c879d4b3a215c189ab1207032dde2dff43440e09aa67bc
-size 1434624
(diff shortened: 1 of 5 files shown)


patmauro:
Working great on the test server so far after the sound fix! :) Looks like this, along with https://github.com/H-uru/Plasma/pull/1462, should partially resolve https://github.com/H-uru/moul-assets/issues/33.

Hoikas (approved):
Thanks for restoring these doors! I took the liberty of changing the museum doors to respect the value of the SDL variable `islmMuseumDoorFunc`. There is no corresponding variable for the library doors, so I left that one alone.

Issue #231, opened 2023-04-19
Bahro Cave: Kadish Pole missing animation


patmauro:
This is an old one, dating all the way back to the original 2003 release - but I figured it'd be worth cataloging here.

In the ABM bahro cave, the pole for Kadish Tolesa does not animate like the other 3 poles - the light pulses in and out, but the "static" animation does not occur; the pole's details are instead frozen in place. However, in Relto this same pole works fine.

![image](https://user-images.githubusercontent.com/6642931/233183284-b1726bf5-e264-41b5-8178-beaf9eecdc4a.png)

patmauro:
Fixed by https://github.com/H-uru/moul-assets/pull/244

PR #244, opened 2023-10-12
Fix Bahro Cave Kadish pole snow animation

DamnBriggsy:
Snow ("static") animation for the Kadish pole in the Bahro Cave did not actually move due to what is likely a copy-paste error. A small change to this prp fixes the issue mentioned in https://github.com/H-uru/moul-assets/issues/231

The Layer Animation `Snow1_LayerAnim_1` had line 10 of the UVW Transform animation controller changed to 
`			<hsMatrix44 identity="True" />`
where previously it was a duplicate of line 7.


Hoikas:
It appears that this PR changes a number of the ObjIDs for textures used in the YeeshaCave PRP file. Can you verify that these references have not been shattered by an old PRP Shop? If you're working with a current build of PRP Shop with all PRPs opened, I would not expect for any ObjIDs to change in this PR.

image

DamnBriggsy:
Looking at bhro_lightsupport*0#0.hsm in BahroCave_District_Textures.prp, its ObjID is 9. When I open BahroCave.age in prpshop and save it, it seems to change BahroCave_District_YeeshaCave.prp to match. If open all 3 .prps individually in prpshop, the ObjID for the same texture is still 9. This happens in the build I have from January and the installed version currently up as the release in the repo.

DamnBriggsy:
YeeshaCave prp was saved first, then the textures prp following a series of events culminating in https://github.com/H-uru/libhsplasma/pull/279. Everything looks okay on my end after this.

Hoikas:
I'm still seeing reference [shattering](https://www.youtube.com/watch?v=3_Y5J0ka4_k)

image

Hoikas:
I've reproduced the changes with PRPShop 2022.05.19 and pushed the result to this PR. I see no evidence of ObjID shattering in my commit. The changes in the GenericPhysicals is expected. Can you make sure the fix still works as expected?

DamnBriggsy:
This looks like it might be identical to another version of this file I created by either opening just YeeshaCave's prp or all 3 without the .age file. Regardless, the visual effect is the same (all textures in the age are scrambled).

The only time it's worked for me has been when the ObjIDs get messed with, and this happens when I open the .age file in prpshop. The .age file needed to be modified to have the nonexistent page removed in order to be opened in this way. 

image

Hoikas:
I reverted my commit to this PR because I see the same breakage. According to libHSPlasma, the changes to ObjIDs made by @DamnBriggsy actually fix incorrect references already in YeeshaCave.prp. What's weird is that the game does not report any problems with the current dataset.

Hoikas (approved):

DamnBriggsy:
Could a maintainer please remove the tag `awaiting author`?

PR #251, opened 2024-03-02
Add checkbox to enable/disable DynamicCamMaps to the Options Menu.

### LINKED ISSUE #247: Need to Remove EAX references from UI
EAX is dead and is not coming back - so even though there is an [ongoing effort to emulate these via EFX](https://github.com/H-uru/Plasma/pull/1000), it would still be prudent to remove the references in the UI to the EAX technology _specifically_, especially given how much of Uru's sound logic has already been ripped out in favor of OpenAL.

Specifically, we need to do the following:
- Remove the EAX graphic at the bottom of the Audio Settings menu.
- Either hide the "Enable EAX" option outright until an EFX solution is ready, or simply replace the string with "Enable EFX," and allow this to similarly be greyed out until an EFX fix is ready.
- Since this is a UI change, the changes should be done via xKIGUI.max.

PR https://github.com/H-uru/moul-assets/pull/239 attempted to solve this issue by replacing the EAX graphic with an OpenAL graphic and by directly modifying `GUI_District_AudioSettingsDialog.prp` to change the "Enable EAX" string to "Enable EFX" - however, the feedback was that (A) we should instead remove the graphic outright, and (B) the needed changes should instead be made in xKIGUI.max and reexported afresh so that the source files can reproduce the game files exactly.

Hoikas:
This updates the options menu to add a checkbox for enabling/disabling DynamicCamMaps. This was possible in Myst V, so it was just a matter of updating the options menu GUI and the options menu script to support this option. I took the verbiage "Dynamic Reflections" from Myst V. Here is the updated graphics settings dialog:

image

In Myst V, it looks like:

image

I removed the subtext for the windowed mode checkbox for the mac client. The TransGaming mac client and its limitations are no longer relevant to us, and the windowed checkbox works fine on the current H'uru mac client.

As drive bys, I updated the localization files to include all of the strings used by the options menu and added references to those into max. This means that the manual localization of the dialog can be removed from xOptionsMenu.py. Some strings are in weird places, such as audio strings being in the game set - I just rolled with what was already there to reduce noisy deltas. Further, I fixed #247 by removing the EAX logo and resizing the audio settings dialog to remove the deadspace formerly occupied by the EAX logo. User facing references to "EAX" have been replaced with "EFX". The audio settings dialog looks like this now:

image

Note that this requires a corresponding update to xOptionsMenu.py, which is forthcoming.

NOTE: Due to H-uru/korman#407, using this checkbox with the K'veer PRPs currently in this repo seems to have no effect. This is by design in the engine. Once the cited issue is fixed and K'veer is re-exported, everything should work correctly.


Deledrius (approved):

Hoikas:
As of #253, this checkbox causes the DCM in K'veer to turn on/off as expected.

PR #258, opened 2024-09-23
Add Fan01 clothing source files

### LINKED ISSUE #257: Version control the Fan01 clothing sources.
The max files for GlobalClothing_District_[Fem|M]ale01.prp were uploaded to the OU discord today https://discord.com/channels/282045216221822978/863924968747958333/1287513197368316008. They should be added to the appropriate place in the repository. Uploading zip here in the interim.

[GlobalClothing.Chloe.20131203.zip](https://github.com/user-attachments/files/17091612/GlobalClothing.Chloe.20131203.zip)

dpogue:
I have not tested re-exporting these, which is probably worth doing... I also put these in the existing clothing folder with the existing 3DS Max files instead of a separate Fan01 folder, but I can do that if we prefer that option.

Closes #257.


Hoikas:
Let's go ahead and get this in.

PR #259, opened 2024-09-23
Add Explorer's Emporium reward clothing

dpogue:
Also involved adding an icon to the existing Pub reward jacket in the (fe)male04 files.


dpogue:
Actually going to close this and let Doobes open a PR with all the clothing updates and source files and compiled files that have been round-tripped and confirmed to work. Doing this in multiple pieces is too easy to get weird keyring mismatches

PR #234, opened 2023-04-23
Add contributing guide.

Hoikas:
This repository exists in an odd state where it kinda matches MOULa but not really. So, I've written up a draft contributor guide that lays out where I think this repo should fall in the ecosystem. Specifically, we are not trying to be the MOULa repository - that's not been our goal, ever really. Our goal, generally speaking, has been to provide a high quality baseline for any shard to operate off of.

One of our main sources of difficulty is that a lot of things that have been tested on Minkata and deployed to MOULa. Unfortunately, Minkata is a testing environment, not a review environment. This has led to content going to Cyan's shard that, even in cases going back almost a decade, isn't particularly up to snuff with the rest of the game. This has been accelerating over the last couple of years and has been putting us in an awkward position where we have a lot of open pull requests whose purpose is basically to match what's on MOULa but not to actually receive feedback.

So, I've codified that we aren't trying to match MOULa exactly (but we also don't want to break it), and we also don't want to be the initial submission point for new Ages. There still isn't an official process, but this at least draws the line in the sand, so to speak, that we expect for content to go through some kind of review process. Ideally, this will be with community members that we know have a keen eye for details both in regard to canon and art. I'm not sure we should explicitly list those people out, so I labelled them as "discerning members" in the document.

With this guide in place, we will need to re-evaluate all fan content that we've accepted and determine whether or not it falls in line with the new guidelines. My initial observations are:
- Chiso: ***Accept*** - with the MOULa roleplay items from the lower floor removed from this repo. Could probaby be moved into a new PRP that we just simply don't include.
  - Endorsed by Hoikas
  - Provides important gameplay service of library Age
- Highgarden ***Revert***
  - Lacks an endorsement
- GoMePubNew ***Accept***
  - Endorsed by Hoikas... and possibly two others from the Gehn era
  - Provides important socialization space, matches the rest of the game content
- Kirel Speeches ***Reject***
  - violated specific shard roleplay rule 
- Serene ***Revert***
  - Lacks an endorsement
- Tiam ***Revert***
  - Lacks an endorsement
- trebivdil ***TBD***
  - is winning the RAD a sufficient endorsement?
-  Veelay ***Accept***
   - Endorsed by Hoikas... and possibly two others from the Gehn era
   - Provides important community space
- Vothol ***TBD***
   - is winning the DLC a sufficient endorsement?

[...]

PR #261, opened 2024-11-02
Improve Green Storm Texture

patmauro:
Improves the variant of the stormy cloud texture Relto uses when the while the "green sky" (Veelay) page is active.

This is the counterpart to [content/145](https://foundry.openuru.org/gitblit/tickets/?r=MOULa-current-content.git&h=145) in the OpenUru Foundry, included in the Q3 2024 update to MOULa. However, there are some notable differences to be aware of, due to OU/MOULa and H'uru/moul-assets being slightly out of sync:
1. OU contains an optimization made by @Hazado which splits out all the fan content into separate PRPs so they are more modular, but this change is not present in H'uru/moul-asets yet. As such, the equivalent change here is applied directly to **Personal_District_Textures** rather than **Personal_District_psnlFanYeeshaPages01**, as the asset in question has not been split off from the _Textures file and repackaged into _psnlFanYeeshaPages01 in the version currently on H'uru/moul-assets.
2. There is a known issue where the sun and moon objects do not display properly through the cloud layer, and this is especially noticeable here. @Hazado added a fix which is currently on OU/MOULa adjusting the layer the objects are drawn in which fixes this issue, but H'uru/moul-assets does not have this fix yet. As such the sun and moon will not be visible while this "stormy veelay" variant is active.

### DIFF
--- a/compiled/dat/Personal_District_Textures.prp
+++ b/compiled/dat/Personal_District_Textures.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:4faabd8cb957d93a225cd60054b6229a3dbc83e47440fb193d1e7f35859e56a7
-size 45236117
+oid sha256:950a5ddc50a4a38568359568b83adae888d2bc0eda8b6266364b755942ae69b4
+size 45236133
--- a/sources/textures/tga/Personal/psnlnewsky3verydark_40#2.png
+++ b/sources/textures/tga/Personal/psnlnewsky3verydark_40#2.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:dd12b061e74abea11f096ce94e3b82979299a8728a9c975680b0ebf76880dc24
+size 283002


Hoikas (commented):
Have you added or updated the texture source file in the sources/textures directory?

Hoikas:
As a more general comment - we haven't been accepting much *new* fan content into this repository as of late, mostly due to lack of review manpower. The sheer volume of activity these last few days has drawn my reviewer's eye, so I'm trying to look at basically everything. I'm wondering if, as a stratagem, we should simply back out all fan Ages and storylines and content associated with those fan Ages. In that way, the focus of this repository becomes a baseline for building a generic shard that can fairly easily have new Ages grafted onto it. Anyway, those musing don't IMO impact the fact that this PR is likely an improvement upon what is currently in master and will probably be accepted before any such mission change happens.

patmauro:
> Have you added or updated the texture source file in the sources/textures directory?

Sure, I can do that - will update "sources" accordingly when I get a sec.

dgelessus:
Regarding @Hoikas' comment - for reference, the last discussion about what amount/kind of fan content should be kept in this repo: #234

patmauro:
> Have you added or updated the texture source file in the sources/textures directory?

Hey Hoikas - I'm not seeing an original version of psnlNewSky3VeryDark anywhere in our source textures, nor the variant of this that was previously created to cover the "stormy sky + veelay sky" case, so it doesn't seem like there's anything to replace. I can add what I have, but without the original texture present, it doesn't seem like there will be much point.

Hoikas:
I think adding your new version would be an improvement on not having anything at all :wink:. When the relto page was originally accepted, we were probably trying to move too fast, not reviewing things appropriately just because they had been accepted to Cyan's shard.

patmauro:
> I think adding your new version would be an improvement on not having anything at all 😉. When the relto page was originally accepted, we were probably trying to move too fast, not reviewing things appropriately just because they had been accepted to Cyan's shard.

Fair enough :) we should try and see if we can track down Cyan's original storm cloud texture - could have sworn that was floating around somewhere - but for now I have added my green variant to the source textures under "Personal" using the same name as the reference in the prp.

Hoikas (approved):
This is a clear improvement. Thanks!

PR #262, opened 2024-11-02
Fix Link Panel for Rainy Cleft

patmauro:
Equivalent to [Commit 3df27e](https://foundry.openuru.org/gitblit/commit/?r=MOULa-current-content.git&h=3df27e5988054d2a4083bacca88974c9c859e456) of [content/144 ("Cleft Revamp Bundle 1")](https://foundry.openuru.org/gitblit/tickets/?r=MOULa-current-content.git&h=144) in the OpenUru Foundry, included in the Q3 2024 update to MOULa. (For submission to H'uru I am splitting this bundle into three separate PRs for increased modularity.)

Fixes an aesthetic issue introduced by MOUL: When Classic Uru was updated for MOUL, almost all linking panels in the game received a graphical overhaul with the notable exception of the panel for Rainy Cleft, which did _not_ change even though Sunny Cleft's panel did. As a result, the panels no longer mirror each other like they did in the base Uru game. This fix addresses this by introducing a new panel for Rainy Cleft that more properly mirrors the updated panel for Sunny Cleft.

Before
image

After
image


dpogue:
Out of curiosity, was this taken from a direct screenshot in game, or was it brightened before being added as a texture? It seems significantly darker than the original (and while yes, some of Cyan's artificially lightened linking panels are brightened to the point of being washed out, it seems like some middle ground might be possible with this particular case)

I do like matching the perspective :+1:

patmauro:
> Out of curiosity, was this taken from a direct screenshot in game, or was it brightened before being added as a texture? It seems significantly darker than the original (and while yes, some of Cyan's artificially lightened linking panels are brightened to the point of being washed out, it seems like some middle ground might be possible with this particular case)
> 
> I do like matching the perspective 👍

I did touch this up _slightly_, but I honestly prefer the much clearer contrast this offers. In addition, the response from those who tested this in-game while preparing the Q3 update has been extremely positive; had I received any feedback asking to adjust this more I might have taken another look but at this time I'm happy with it where it is.

Hoikas (commented):
Have you added the new texture to sources/textures?

patmauro:
> Have you added the new texture to sources/textures?

I can do that - would you prefer I _overwrite_ xLinkPanelTomahnaDesert.tga or add a new file, ie xLinkPanelTomahnaDesert2 or something? I'd be more comfortable going with the second option and keeping both, but your call.

Hoikas:
Overwrite the original file or delete it and add a PNG. The old version will still be available in git history.

patmauro:
> Overwrite the original file or delete it and add a PNG. The old version will still be available in git history.

Done. Let me know if any issues.

Hoikas (approved):

PR #263, opened 2024-11-02
Cleft Texture Improvements (Zandi Props & Skybox)

patmauro:
Equivalent to [Commit 0a34a4](https://foundry.openuru.org/gitblit/commit/?r=MOULa-current-content.git&h=0a34a4d81595ec5d08e70504269a55b2a3a1cec2) of [content/144 ("Cleft Revamp Bundle 1")](https://foundry.openuru.org/gitblit/tickets/?r=MOULa-current-content.git&h=144) in the OpenUru Foundry, included in the Q3 2024 update to MOULa. (For submission to H'uru I am splitting this bundle into three separate PRs for increased modularity.)

First batch of a planned series of improvements to help modernize The Cleft:
- Gives "ClftBeerCan" a facelift that reads more clearly as a water bottle and makes sure the "Sponakee" joke is actually legible.
- Slightly bumps up the res on ClftCD so this reads more clearly.
- In keeping with MOUL's intended "current day" setting, gives ClftChipsBag a "modernized" redesign (created entirely from scratch with 100% original assets) so Zandi does not appear to be eating from a 30-year-old bag.
- Slightly bumps up the res on ClftRadioMap (but only _slightly_ - the full res we have available in the original assets is _far_ larger than is necessary.)

In addition: The Cleft's skybox appears noticeably lower-res compared to other ages, which is unfortunate as this is a very important location. To address this, this change bumps up the res on ClftSky3 to the original texture resolution (2x). (Note: @Hoikas and I investigated various compression methods on this and were unable to find one that did not introduce any noticeable artifacts to the in-game sky - so I can understand if that makes this particular change controversial/unacceptable for H'uru/moul-assets.)

Note: I assessed several other objects in my tests as well, such as the chair, table grain, grill, cooler, etc, but did not find that there was any noticeable or worthwhile improvement in these cases - the set I've prepared here seem to be the only of Zandi's props that actually benefit from TLC.

image
image

image
image

### DIFF
--- a/compiled/dat/Cleft_District_Textures.prp
+++ b/compiled/dat/Cleft_District_Textures.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fceec4a3b884c772e559c0f73e8c2e534ce0c9c66c3b7dc0e8c12a89c4586727
-size 35044682
+oid sha256:14e5121930a9c70e7991ab2f632b7b2000b61ba45317f1769cf8cbfddc694390
+size 37242158
--- a/sources/textures/tga/Cleft/clftBeerCan.png
+++ b/sources/textures/tga/Cleft/clftBeerCan.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:cb2a1aad9ad6ddae2877760b1240e459cebe3e705c6e5d01b957098589d31920
+size 26580
--- a/sources/textures/tga/Cleft/clftBeerCan.tga
+++ b/sources/textures/tga/Cleft/clftBeerCan.tga
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:21b5f4014187406bc24c297a206a0f2958010a2802e5d8cb4804d1cc1244972a
-size 65580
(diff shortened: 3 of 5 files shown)


Hoikas (commented):
Can you update or add the new textures to the textures directory in this repository?

patmauro:
> Can you update or add the new textures to the textures directory in this repository?

I can do that - CD, radio and sky are already good, but chips bag is new, and water bottle has been altered enough that it may be worth saving in the textures as well. I will add these to "sources/textures/tga/Cleft" as "clftBeerCanNew" and "clftChipsBagNew" if that sounds good.

Hoikas:
Sure. If the old version is already in the repo, you can just replace it. The old version will still be available in git history.

patmauro:
> Sure. If the old version is already in the repo, you can just replace it. The old version will still be available in git history.

Done. Let me know if any issues.

Hoikas (approved):
The beer can's increase in resolution by 4x is maybe a little much, but I think the Textures PRP and VRAM budget are still respectable.

patmauro:
> The beer can's increase in resolution by 4x is maybe a little much, but I think the Textures PRP and VRAM budget are still respectable.

Yes, this was definitely the outlier here - not sure if it has to do with the mesh or whatever, but that's the smallest I could get it and keep it still legible. Needless to say I doubt I'll be making any further changes like this one anytime soon. Thanks for taking a chance on this one though - I think making sure the little jokes Cyan put in are legible is a net positive for Uru.

Hoikas:
S3TC's block compression algorithm is a pain for that, and I don't know that we're ever going to be able to use anything more modern, given the pushback we get on even more modest tech upgrades.

PR #266, opened 2024-11-02
FemaleHandsOptionsColorsFix

TheScarFr:
Fix the Female avatar's Hands Options colors switch when selecting a different Hands Option:

The fix constist to put the Nails color in the layer 1 and to put the Gloves color in the layer 2,
for both the Fingerless Gloves and the Canvas Gloves.
It have been done in the "sources\avatar\clothing\bas.avatar.female.max" file with "3DS Max 2022", and then the "compiled\dat\GlobalClothing_District_Female.prp" file have been exported from the MAX file.

### DIFF
--- a/compiled/dat/GlobalClothing_District_Female.prp
+++ b/compiled/dat/GlobalClothing_District_Female.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:eddd63e47d6dc28a1e6b669f55be7d3dd825ce0f9bfe265f075e0e7139a5b9ea
-size 4731529
+oid sha256:50a9a17154035b8227a38e9ace222b033c5add70481ddde0ea0010cbb945f0ed
+size 4731704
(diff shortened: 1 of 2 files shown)


Hoikas:
> Fix the Female avatar's Hands Options colors switch when selecting a different Hands Option:

I'm not sure I understand what you're trying to fix based on this description.

dpogue:
> I'm not sure I understand what you're trying to fix based on this description.

If you set a fingernail colour for a female avatar, and then put gloves on, and then remove them, it loses the nail colour due to different clothing items using different tint layers for the nails rather than being consistent. The fix is to consistently use the same tint field for the nails

EhrenCG:
Dang it, you explained before I could dpogue lol, but I did create this image of the effect of how the color swaps when you just swap the item by using a red and cyanish color here:

I still like visual representations in general so I'm still posting it. (This is obviously _without_ the fix.)

MovingNailColor

Hoikas (approved):
Thanks for the fix!

TheScarFr:
`[update] Wardrobe: Fix for the Female Hands Options colors. {TheScar(.fr)}`

PR #268, opened 2024-11-17
Make the Nexus ***dynamic***.

Hoikas:
This removes the old per-Age linking panel objects from the Nexus max file. Instead, I've exposed the main book's linking panel layer to Python via the new `ptAttribLayer` and expect for the Python script to swap the texture out dynamically. This should help us become more independent from fan-created Ages.

There is still some work to be done here:
- Most of the city linking panels are not in GUI_BkBookImages, so they appear as the black void for now.
- There are still a stray linking panels in the Nexus PRP - it looks like they are the Guild Pubs.
- The duplicate targas in the Nexus directory need to be disposed of without causing the max file to explode.

NOTE: To export this, you need
- H-uru/Plasma#1631 (required)
- H-uru/Plasma#1634 (optional, recommended)

To play, use H-uru/Plasma#1633


Hoikas:
I've force-pushed a fix for the wrong PRPs being included. Whoops! This is what happens when you juggle lots of branches and don't test before pushing.

Also, given that there is still some work to be done on this PR, I'm going to mark it as a draft until those tasks are completed.

Hoikas:
All city linking panels should now be working. I still need to pull the Guild Pub panels out of the Nexus' textures PRP and import them into BkBookImages. Further, there are now missing image warnings that need to be squashed in the Nexus max file.

Hoikas:
This is ready for prime time

dpogue (approved):

PR #270, opened 2024-11-18
NoLinkOutGlobalAnimation

TheScarFr:
Add the "NoLinkOut" GlobalAnimations:

I have 2 PRPs for a new GlobalAnimations (`GlobalAnimations_District_[Gender]NoLinkOut.prp`) which I'll need for my "Trigon" Age, but this Age isn't ready yet; that being said, as these animations are global ones, someone could use them in his own Age.

These animations play forward the "LinkOut" animations without the final loop, keep the avatar with his hand raised for about 2 seconds, and play backward the "LinkOut" animations (still without the loop).
Here is a vidéo of the animations: https://youtu.be/_UdFXcMbWjE

They look a little like the "InsertKiHandLonger" animations, but with the right hand instead of the left hand.
It could be used in-game for different things.

Changes: `GlobalAnimations_District_FemaleNoLinkOut.prp` and `GlobalAnimations_District_MaleNoLinkOut.prp` new files.

- Development for POTS: `Sirius` > His email address isn't written on his profile. ( @Jrius )
- Convertion for MOUL & Edit: `TheScar(.fr) <thecleftbeyondthescar@gmail.com>` ( @TheScarFr )
- Convertion fixes: `Ehren <EhrenCG@hotmail.com>` ( @EhrenCG )

See also: https://github.com/H-uru/Plasma/pull/1636

### DIFF
--- a/compiled/dat/GlobalAnimations_District_FemaleNoLinkOut.prp
+++ b/compiled/dat/GlobalAnimations_District_FemaleNoLinkOut.prp
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:ad782d545e5e77b44edd3704c956308b9ff4241c867d19feab07f981354f212d
+size 30286
--- a/compiled/dat/GlobalAnimations_District_MaleNoLinkOut.prp
+++ b/compiled/dat/GlobalAnimations_District_MaleNoLinkOut.prp
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:c482f90ba36217f40f5dd864bf5e9ef25a6861f42741611b4370c80461ef325d
+size 66038


Deledrius:
What are the situations in which this animation would be useful?  Putting your hand on a linking panel but not going anywhere shouldn't ever happen.  I don't think it's a good idea to put new animations in the global space unless they're likely to be useful in a wide range of Ages.

As for the animation itself, it feels unnatural to have it reverse the first half of the motion.  It doesn't look like how a person would relax from having their hand and lowered head in that position to me.  If someone wanted to simply reverse an existing animation, I believe they can do that with a behavior without needing a whole new animation to do it, making this redundant.

TheScarFr:
I'll need these animations for the "Trigon Access Bracelet" dispenser, in the "Trigon" Age. ( https://youtu.be/nlZmEEUNoY4 )
The other animations where to short, or where with the left hand; the bracelet is on the right hand.
I asked many times how to use the Multistage Behavior but never get any answer; but I'm not sure it would have work with my dispenser anyway...
So, Sirius wrote a script to merge the forward and backward anims without the loops and with the 2 seconds raised hand.

dpogue:
2 thoughts:

* `NoLinkOut` is a bad name for an animation because it doesn't tell you anything about what the animation does. Something like `InsertRightHand` would be much better.
* Age-specific animations are probably better off starting as animations within that Age, before being promoted to global animations when needed.

TheScarFr:
I chosed the "NoLinkOut" name because the anims are made from the "LinkOut" anims, but of course I can change it to "InsertRightHand" (I did already thought about choosing it).
Age-specific animations shouldn't have existed and all the animations should have been Global ones, IMO; it uselessly complicates the things. /shrug

dpogue:
I don't think that an animation used only for a single Age is going to get accepted here as a global animation. One reason is that it increases the amount of memory needed to run the game (because all global animations are loaded at startup) for no benefit, rather than only being loaded when necessary for the specific Age.

If there is demand to reuse animations in multiple Ages, then we can evaluate promoting it to a global animation, but it's too early to make those assumptions now (when this animation is currently unused anywhere in the game).

TheScarFr:
I understand that, but how people could want to use these animations if they are hosted nowhere?
It's the same paradox as the chicken and the egg... Who was there first?

Deledrius:
If someone makes an Age that uses an animation which only exists in another Age, then they either copy it to their new Age or request it be made global at that point.  That's not really a chicken-and-egg issue, just a policy of only putting things in global once they have multiple users.

patmauro:
Scar - something else to consider is that by keeping this bundled as part of Trigon itself, you'll help ensure that Trigon will be more modular by reducing the amount of potential dependencies the Age has on the base shard's content. This will help keep your age content cleaner and will ensure you will be able to test/deploy in a variety of different environments. Plus, as others have said, if this _does_ end up being a resource lots of people want to implement, then there's no reason you couldn't explore converting this to Global at some point in the future. So I think keeping this animation Trigon-specific is probably a win-win for you.

Deledrius:
An excellent point, @patmauro!  That's important for testing and deployment.

TheScarFr:
I understand that, but do you understand that if the Trigon Age isn't ready yet, and that if the anim is hosted nowhere, nobody can use it in his Age?

Deledrius, I wasn't talking about the fact that someone adds in his Age the anim from another Age, but about the ability for him to know that this anim exists. If it's added nowhere, how can he know that it exists?

patmauro:
> I understand that, but do you understand that if the Trigon Age isn't ready yet, and that if the anim is hosted nowhere, nobody can use it in his Age?

Yes, I do understand that. But like I said - if that becomes an issue, and enough people ask to use this animation in their Ages in the future, then you can always look into converting this to Global at that time, so there's no risk leaving this out of Global until you know for sure. So: there's really no harm in keeping this bundled with Trigon for now - it'll actually help make sure your Age is modular and easy to test in different environments. But again, you can always re-assess if there's enough demand in the future - so like I said, it's a win-win scenario for you.

TheScarFr:
I'm not sure to understand what you think about when you say "different environments" (language barrier?).

TheScarFr:
I'm aiming only MOUL(a) so also Minkata (and Complete Chronicles) for my Ages, but anybody could use this anim somewhere else.

PR #271, opened 2024-11-18
Adjust Translations for "Big Blues" Attributions

patmauro:
Very minor change: adjusts the translations for the "Big Blues" attributions using translations confirmed by fluent speakers who have been co-authored. Once approved, I will copy over this change & the other adjustments we made as part of #264 to OU for consistency.


Hoikas (approved):
Thanks for the improvement!

PR #260, opened 2024-09-24
Add Riven 2024, Radio Free D'ni, ExpEmp, and Mysterium 2024 Clothing

DoobesURU:
* Add Riven 2024 shirt and sources (dpogue)
* Add Radio Free D'ni shirt (long and short sleeve) and sources (AlanD, et al.)
* Add Explorers' Emporium reward tunic and sources (Semjay)
* Add Mysterium 2024 shirt and sources (Doobes)
* Update Fan02 and Mysterium JSON files with new clothing definitions

PR #276, opened 2025-02-22
Clothing 2025-06

dpogue:
Supersedes https://github.com/H-uru/moul-assets/pull/260 (which broke because of changes to the Female base clothing that were merged)

- [x] Update the Fan02 clothing JSON file to have page references
- [x] Update the Fan02 clothing PRPs
- [x] Update the Mysterium JSON file to have more page references
- [x] Update the Mysterium clothing PRPs
- [x] Figure out what (if anything) needs to be done for the Pub Jacket PRPs


Hoikas:
Do you think this will address #283?

dpogue:
> Do you think this will address #283?

No, I believe that's an actual missing icon (in the original Cyan PRPs, because previously it didn't need one)

Hoikas (approved):

PR #284, opened 2025-07-04
Fix two more options menu text boxes not using loc files.

Hoikas:
Supersedes #280 by fixing the issue in the Max file.


dpogue (approved):
I've tested this in-game and checked in PrpShop and it looks good to me

PR #286, opened 2025-07-06
Add state for smashed Bevin heek table

dpogue:
This is missing the sound emitters and holographic sprites from the offline game (but they could be restored if there are strong feelings about them). It seemed in-character that a table smashed in February 2004 might have stopped working in the intervening 20+ years.


Hoikas:
Those sprites and sound emitters were a pretty nice touch in the original smashed tables.

dpogue:
yeah, they didn't import and re-export very well with ZLZ, but I'll see what I can do about reviving them.

dpogue:
I poked a bit more at this, and ZLZ is not happy with importing the multiple chained LayerAnims for the sprites, and I've not been able to get them sorted out in Blender or by hand patching PRC objects.

I suspect the easiest way to do this is going to be some sort of PRP splicing "object extractor into new page" tool (which is a weird definition of "easier").

I was looking at Gnasty Plasma Patcher and it looks like it has the core bits of what I'd need, but I didn't poke deep enough to see if I can give it just a subset of objects to extract/convert

PR #281, opened 2025-05-10
GreenBushesReltoPageKveer

TheScarFr:
Add a "Green Bushes" Relto Page in K'veer.
It adds back the Green Bushes from the Complete Chronicles' "Lush" Relto Page, but in a new Relto Page in K'veer, where was the Complete Chronicles' Lush Relto Page.
Related Pull Request from Plasma: https://github.com/H-uru/Plasma/pull/1693

Changes:
```
- compiled\dat\
	- GUI_District_BkBookImages.prp: Exported from "sources\textures\tga\GUI\GUI_BkBookImages.json".
	- GUI_District_YeeshaPageGUI.prp:
		- Coordinate Interface:
			- YeeshaPage32-GreenBushes
			- YeeshaPagePlane:
				- YeeshaPage32-GreenBushes (Scene Object)
		- Draw Interface:
			- YeeshaPage32-GreenBushes:
				- Drawables:
					- GUI_YeeshaPageGUI_10000008_0BlendSpans4 (Drawable Spans)
		- Drawable Spans:
			- GUI_YeeshaPageGUI_10000008_0BlendSpans4
		- GUI Button Mod:
			- GUIYeeshaPageBtn32:
				- Animation:
					- Map #733029_anim (Layer Animation)
		- GUI Dialog Mod:
			- YeeshaPageGUI:
				- GUIYeeshaPageBtn32 (GUI Button Mod)
		- Layer:
			- Map #733029:
				- Texture: xYeeshaPageGreenBushes.dds
		- Layer Animation:
			- Map #733029_anim
		- Material:
			- sketchGreenBushes:
				- Layers:
					- Map #733029_anim (Layer Animation)
		- Mipmap:
			- xYeeshaPageGreenBushes.dds
		- Scene Node:
			- GUI_YeeshaPageGUI:
				- Scene Objects:
					- YeeshaPage32-GreenBushes (Scene Object)
		- Scene Objects:
			- YeeshaPage32-GreenBushes:
				- Modifiers:
					- GUIYeeshaPageBtn32 (GUI Button Mod)
	- Kveer_District_YeeshaPage32.prp: NEW
	- Personal_District_psnlFanYeeshaPages07.prp: NEW
- sources\
	- Kveer\
		- Kveer_District_YeeshaPage32.blend: NEW
	- textures\tga\GUI\
		- xYeeshaPageGreenBushes.jpg: NEW
		- xYeeshaPageGreenBushesAlpha.jpg: NEW
```

PR #287, opened 2025-09-08
Add BkBookImages JSON.

Hoikas:
This adds a JSON file and all of the source images required to rebuild GUI_District_BkBookImages.prp using my [MakeImageLibPRP.py script](https://github.com/Hoikas/moul-utils/blob/master/MakeImageLibPRP.py). The source images were all ripped from the previous BkBookImages.prp and compared with the files already in the sources directory. If an identical lossless version already existed, the lossy version from the PRP was discarded. Otherwise, the lossy version from the PRP has been added here and is re-imported to the the new PRP.

The linking panels are primarily the trouble spots here - they were mostly reshot for MOUL, but what we have here is mostly the TPotS linking panel images. Some of them are identical. Many of the Yeesha journey Ages use the same angle but have been brightened somewhat. As stated above, I brought in the MOUL reshots if there was any discernible difference. No effort was made to correct problems with any linking panels. So, for example, the lower pellet cave linking panel still has a cursor in it.

The resuling diff on BkBookImages.prp should be minimal. There were some RLE changes reevaluated, and any PNGs in BkBookImages.prp were compressed to JPEG. In my work, I discovered using lossless compression resulted in a 60 MB BkBookImages.prp, which is undesirable.

I also deleted all of the TPotS panels that did not match the MOUL variants. Once this is merged, #117 and #281 can be more trivially rebased.

### DIFF
--- a/compiled/dat/GUI_District_BkBookImages.prp
+++ b/compiled/dat/GUI_District_BkBookImages.prp
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:2d4376654f4f8cac12931a3cac4cab4f1cb15ebe73117deacfa4d4d9b3e837d5
-size 10572408
+oid sha256:091bf3f3a2ca0c81cabe84dfdea4aec09e80907a5a3ba3720747c0ea335a0d17
+size 7830053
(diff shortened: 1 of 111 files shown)


dpogue (approved):
It's hard to review this without actually seeing any of the image diffs, but the work to automate generating the BkBookImages file is definitely needed

PR #288, opened 2026-02-21
Revert PR #287 to fix corrupted textures

dgelessus:
The changes in #287 were performed with a buggy version of libHSPlasma that corrupted the alpha channels of mipmaps with a JPEG-compressed alpha channel (see H-uru/libhsplasma#310). Affected textures include multiple Relto page symbols (xYeeshaPageAlphaSketch...) and the Bahro stone share symbol (xBahroYeeshaShare). These textures were corrupted both in the PRP (and thus in-game) and in the extracted image files.

As a temporary fix, this reverts that PR. It should be safe to redo the PR using a libHSPlasma version that isn't affected by this bug.

In-game examples of the corrupted textures introduced by #287:

Bahro stone share symbol, incorrectly showing with a rock-textured square border around it
Relto book symbols for the clock and birds Relto pages, barely visible due to being almost transparent, and with no visible difference in color between the active (left) and inactive (right) symbols

and how they're supposed to look (which is how they look with this PR):

Bahro stone share symbol, displaying correctly
Relto book symbols for the clock and birds Relto pages, displaying correctly and with a clear difference in color between the active (left) and inactive (right) symbols


dgelessus:
Unrelated to the JPEG alpha corruption issue, I just noticed another problem with #287: some of the Relto page symbols got mixed up. This seems to be because the Relto page numbers in Cyan's released textures don't all match the numbers in the MOULa data files. For example, "xyeeshapagealphasketch15*1#0.hsm" in MOULa's data files is the bench page symbol, but "xYeeshaPageAlphaSketch15.tga" from Cyan's textures is the clock page symbol. The JSON file added in #287 incorrectly assumes that the numbers are the same in both, and so with the rebuilt PRP from that PR, the bench page incorrectly shows the clock symbol.

Thankfully, that should be easy to fix by renaming the texture files in the repo and/or adjusting the JSON file to use the correct names.

PR #289, opened 2026-03-09
Fix issues with BkBookImages introduced in #287.

Hoikas:
This fixes some issues with BkBookImages introduced by #287. I re-extracted the previous BkBookImages with a freshly built libHSPlasma from tag 2025.11.04 and copied the resulting PNGs over. I didn't bother with JPEGs because the JPEGs were just dumps directly from the PRP and should have been unaffected. I also corrected the bahro share symbol missing its alpha jpeg and removed some conflicting targas in favor of the correct version from the old BkBookImages.

These changes should make every image in BkBookImages "correct," and this should be a good baseline moving forward.

CC @dgelessus


sources/textures/tga/GUI/xBahroYeeshaShare.jpg
dgelessus, line 1: The resolution of this file went down from 512x512 to 256x256, resulting in a smol share icon in-game:

Bahro stone share symbol, displaying at half its intended size

dgelessus (changes requested):
Thanks for looking at this again. The Relto page symbol textures look good again, but there is still an issue with the Bahro stone share symbol (see below).

It would be great if we could somehow visually confirm that the textures match up between the original PRP and this version, without manually checking each one in-game or in PrpShop...

sources/textures/tga/GUI/xBahroYeeshaShare.jpg
Hoikas: Ah, I see what happened. I pulled this copy of the texture from the initial commit of BkBookImages, but it was enlarged at some point. I will get the correct one in tomorrow.

Hoikas: Image

Should be fixed now

dgelessus (changes requested):
I used a patched version of your [ExtractTexturesPRP.py](https://github.com/Hoikas/moul-utils/blob/4c0bdf629147124851ddf962a8e3c2a66806654e/ExtractTexturesPRP.py) script to compare the previous BkBookImages.prp (from *before* #287) with the current state of this PR.

<details>
<summary>Patch for ExtractTexturesPRP.py</summary>

```patch
diff --git a/ExtractTexturesPRP.py b/ExtractTexturesPRP.py
--- a/ExtractTexturesPRP.py
+++ b/ExtractTexturesPRP.py
@@ -80,22 +80,15 @@
             dds.write(fs)
     elif image.compressionType == plBitmap.kJPEGCompression:
         logging.debug(f"{image.isImageJPEG()=}, {image.isAlphaJPEG()=}")
-        if image.isImageJPEG() and image.isAlphaJPEG():
-            _export_jpeg(output_path, image_name, image, image.jpegImage)
-            _export_jpeg(output_path, f"ALPHA_{image_name}", image, image.jpegAlpha)
-        elif image.isImageJPEG() and not image.isAlphaJPEG():
-            _export_jpeg(output_path, image_name, image, image.jpegImage)
-            _export_png(output_path, f"ALPHA_{image_name}", image, image.extractAlphaData())
-        elif not image.isImageJPEG() and image.isAlphaJPEG():
-            # NOTE: for some reason, plMipmap.extractColorData() is returning garbage,
-            # so we'll just output the decompressed alpha data into the PNG. That may make
-            # the most sense for now.
-            _export_png(output_path, image_name, image, image.getLevel(0))
-            _export_jpeg(output_path, f"ALPHA_{image_name}", image, image.jpegAlpha)
-        elif not image.isImageJPEG() and not image.isAlphaJPEG():
-            _export_png(output_path, image_name, image, image.getLevel(0))
+        if image.isImageJPEG():
+            _export_jpeg(output_path, f"{image_name}__COLOR", image, image.jpegImage)
+        else:
+            _export_png(output_path, f"{image_name}__COLOR", image, image.extractColorData())
+
+        if image.isAlphaJPEG():
+            _export_jpeg(output_path, f"{image_name}__ALPHA", image, image.jpegAlpha)
         else:
-            raise RuntimeError
+            _export_png(output_path, f"{image_name}__ALPHA", image, image.extractAlphaData())
     else:
         _export_png(output_path, image_name, image, image.getLevel(0))
 
```
</details>

I noticed a few more oddities still present. (I *think* all of them are from #287 and not newly introduced in this fix PR, but I haven't checked systematically...)

* `xbookmarkdnizero` and `xbookmarkdnizero2` in this PR have transparency, whereas before they had no alpha channel and instead had a book page texture "baked in" as the background. I think this is because the repo already contained source .tga files with these names, but I guess those are the source textures *before* they had the background baked in. I can't test the in-game effect of this difference, because moul-assets no longer contains any of the ages whose books use this texture... We should restore the previous textures (with no alpha channel) unless someone can confirm that the alpha textures work correctly in-game.
* `xlinkpanelahnonaytemple`, `xlinkpanelblackvoid`, `xlinkpanelconcerthallfoyer`, `xlinkpanelgrtzero` seem to be a bit darker in this PR compared to before. Your commit message for 7fcbac475b6b151005931dd948358d608ed76ea8 mentions this, so it's not surprising. IMO it's noticeable for ahnonaytemple and blackvoid, so I would vote to revert those, but the others are close enough that we probably don't need to care.
* `xlinkpaneltomahnadesert`, `xurucreditslegallogos`, `xyeeshabookstampvsquish` lost their alpha channels. I haven't confirmed these in-game (and at least the credits logos are unused since #61, but still). These should be fixed (or removed entirely, if we know for sure that a texture is unused).
* `xYeeshaPageVeelay7`'s alpha JPEG was previously orange, somehow. #287 changed it to the usual red. This change is completely fine - I'm just surprised by the previous state!

Hoikas:
The D'ni zero images behaved a little strangely, so I reverted those.

Mystler:
This seems to fix a few issues (like the broken relto pages and cleft book) i ran into during my H-uru client test playthrough on Destiny.

I don't know if these in particular are a problem with this PR or not but here go the remaining issues I now saw:

### Ahnonay Cathedral book image broke

image

### Various linking books have small black borders

A whole bunch of books have this issue, these are just examples:

image
image
image

Mystler:
Found another one 
image

Mystler:
There also seem to be discrepancies between bookshelf and original book.
While the image for the Ahnonay Cathedral is just gone and linking there is impossible via Relto or the link out in Ahnonay, the book in the Watcher's Pub does display it but without any alpha mask.
image

Mystler:
Oh, this might explain why ahnonay cathedral broke: ahnonay vs ahnnonay

```diff
diff --git a/GUI_District_BkBookImages.prp.old b/GUI_District_BkBookImages.prp.new
index a1a9b10..8ac8df1 100644
--- a/GUI_District_BkBookImages.prp.old
+++ b/GUI_District_BkBookImages.prp.new
@@ -795,9 +798,9 @@ plMipmap xkingyableshanmtkjrnlcover_eng*1#0.hsm : <plKey "<-2|56>[0004]xkingyabl
 	</JPEG>
 </plMipmap>
 
-plMipmap xlinkpanelahnonaytemple*1#0.hsm : <plKey "<-2|56>[0004]xlinkpanelahnonaytemple*1#0.hsm">
+plMipmap xlinkpanelahnnonaytemple*1#0.hsm : <plKey "<-2|56>[0004]xlinkpanelahnnonaytemple*1#0.hsm">
 <plMipmap>
-	<plKey Name="xlinkpanelahnonaytemple*1#0.hsm" Type="plMipmap" Location="-2;56" LocFlag="0x0004" />
+	<plKey Name="xlinkpanelahnnonaytemple*1#0.hsm" Type="plMipmap" Location="-2;56" LocFlag="0x0004" />
 	<BitmapParams PixelSize="32" Space="Direct" Flags="0x0002" />
 	<Compression Type="JPEG" SubType="RGB8888" />
 	<ModTime low="0" high="0" />
```

sources/textures/tga/GUI/GUI_BkBookImages.json
```diff
@@ -230,16 +238,20 @@
...
     "xYeeshaPageAlphaSketchClock.png",
     "xYeeshaPageAlphaSketchFiremarbles.png",
     "xYeeshaPageAlphaSketchFireplace.png",
     "xYeeshaPageAlphaSketchGrass.png",
     "xYeeshaPageAlphaSketchLeaf.png",
     "xYeeshaPageAlphaSketchLushRelto.png",
     "xYeeshaPageAlphaSketchStorm.png",
-    "xLinkPanelAhnonayTemple.tga",
+    {
+      "name": "xlinkpanelahnnonaytemple*1#0.hsm",
```
dpogue:
```suggestion
      "name": "xlinkpanelahnonaytemple*1#0.hsm",
```

Per @Mystler's findings

sources/textures/tga/GUI/GUI_BkBookImages.json
Mystler:
```suggestion
    {
      "name": "xlinkpanelgomepubnew*1#0.hsm",
      "color": "xLinkPanelGoMePubNew.tga",
      "alpha": "xLinkPanelAlpha.png"
    },
```

Mystler:
After checking with Hoikas, the preference was to use a single alpha file rather than grab all the different ones, so in order to fix the black borders, I manually got the intersection area of all linking panel textures for a new xLinkPanelAlpha mask. I believe this should deal with the pending issues. Please give this another check that nothing else was missed.
cc @dgelessus

dgelessus:
I haven't had the time to re-test/re-review this in more detail and probably won't get around to it very soon, so I appreciate the testing from @Mystler. As this PR has been open for a while and the last state in the repo is known to be pretty broken, I don't mean to block this PR just because I don't have time for a detailed review. As long as we're somewhat confident that this PR is an improvement over the current state of the repo, I think it's okay to merge this. If we find more issues later, we can fix them incrementally.

That said, I wish we could review this more reliably somehow, as we're on the 4th or 5th round of reviews and still finding problems that we missed every time before...

Targeted testing/spot-checking, as we've done before, depends on luck. Systematically testing every changed texture is a lot of effort, because they appear in so many different places in the game. That's why I did a full texture-by-texture diff (outside the game) last time, but even that was quite a bit of effort to look through, because there were a lot of insignificant, visually identical "changes" from the .prp being rebuilt from the source textures. And as we've seen, issues like a slightly mismatched alpha channel are easier to spot in-game than in a series of hundreds of side-by-side diffs.

PR #290, opened 2026-05-13
Mysterium 2026 shirts

dpogue:


dpogue:
This was deployed to MOULa

Hoikas (approved):

Mystler:
I presume this can get merged? It was approved over a month ago and looks fine.
Only caveat here is that this repo currently does not have the latest GoMePubNew files that provide the actually intended way of how to get these clothing items unlocked.

PR #291, opened 2026-05-13
Mahrn Gestoy clothing robe

dpogue:
This is a clothing item proposed for introduction in the Mahrn Gestoy fan-Age. I don't know how we want to handle that, per our content policies, but I'll open this such that it's available if we want it.

Clothing is probably one of those things where we don't want files drifting too far out of sync because it will cause crashes if someone on the server is wearing clothing the local client doesn't know about.


dgelessus:
Wouldn't it be better to put this in its own, named page? I think I remember a past discussion where someone said that we'd prefer not to put more new clothing into the generic Fan01/Fan02 pages.

I know that we can now easily rebuild/modify these pages using the Python script, but keeping unrelated clothing in separate pages still has advantages. Especially in a case like this, where (as I understand it) the clothing is linked to a specific fan age, shard admins might not want to pull in this clothing item if the shard doesn't include the corresponding age. That would be much easier if the clothing is in separate pages and not part of the generic Fan02 files.

dpogue:
Per the comment from dgelessus and subsequent support from a discussion in the OU Discord, I've separated the MGRobe clothing into its own self-contained PRP files

PR #293, opened 2026-08-19
Move the "disabled items" responder to the correct PRP.

### LINKED ISSUE #1571: Rendering Regression in Options Menu
There is a rendering regression in the options menu from the legacy client to our client. In the options menu, when an option is invalid, such as turning on EAX effects, the option is greyed out. The greying of the text is done by changing the foreground color of the text in Python. Greying out the checkbox is done by animating another greyed out box on top of the ordinary check box. At some point, this greyed out check box has either stopped rendering correctly or the responder controlling the animation has stopped firing as expected. I copied and pasted the GUI prps currently available on Minkata and see this on the audio options dialog:

image

This is the appearance using the same PRPs on Minkata:

image

Hoikas:
The disabled items responder for the options menu was incorrectly placed in the KIYesNo page. This worked fine until the addition of `pfConfirmationMgr` made the KIYesNo page be lazy-loaded instead of loaded alongside all of the rest of the KI. This fixes H-uru/Plasma#1571. There are a few changes in the PRPs that are consequences of the mipmap modification time changes and, apparently, a texture that went amiss at some point.


Mystler (approved):
Diffs look good, intended issue is fixed, and I didn't notice any problems with GUIs or the removed texture when checking various UI elements.

PR #294, opened 2026-08-24
Fix sparkly rendering order in Jalak and Ercana

### LINKED ISSUE #256: Jalak Sparkly rendering order issues
Reported by Keiya in Cyan Discord:
image

Mystler:
Fixes #256 

To accomplish this the sparky was moved to a new drawable span.
- In Jalak, there were only 2 shared icicles in the span, thus it was easy enough for me to cleanly separate the sparky out so that each span has only the buffers and icicles for its actual objects. I recalculated bounds and space trees for each using libhsplasma.
- In Ercana, since the drawable span has 46 objects that all share it, I made the call to not touch the original span at all and consider the fact that it has 4 extra vertices and 6 extra indices in the buffer a non-issue that is not worth re-indexing everything else for. So for Ercana I simply made a new sparky-only span and pointed the draw interface at that instead.

image
image


Hoikas (approved):
I'm not a huge fan of single material drawables because it's less efficient to render, but this isn't eh first time we've accepted those, and it's just a plane, so it's probably not a big deal.

PR #296, opened 2026-08-29
Restore vertex colors in Ahnonay Sphere 2

Mystler:
Can't have this bug get its 20th anniversary, so I'm here just in time to provide the fix.

image

So, what happened? The vertex colors must have gone missing on some export and have never been fixed in MOUL. Fortunately, we do have a reference file that worked back in in the offline game with TPotS/CC. I went about this in two stages:
1. Figure out how to write a simple tool that changes vertex colors at the vertex buffer position that I want to change. libHSPlasma is the ideal choice for a library here. Major thanks to @Hoikas here for being there to bounce ideas off of as well as knowing some of the related quirks of HSPlasma that i needed to work around to get it to mark the changed vertex data as dirty and properly save it to the modified file.
2. Once I had that, try to load the TPotS reference and search its vertices for those matching the ones in the MOUL file and use the old vertex colors. I converted all vertex coordinates into world coordinates in order to find the matching ones. This alone was not enough since since the TPotS version is placed differently in the world than the MOUL version with that offset baked into the vertex positions. Thanks to our favorite object the LinkInPoint, I could easily see that the final coordinates for it in world space are translated by exactly -2000 on the Y axis though, so factoring all of that into the comparison math, I then got exact matches within 0.01 precision per axis for all vertices for the intended color lookup.

This is the uncleaned, undocumented, and read-at-your-own risk hackery that ended up doing this. https://gist.github.com/Mystler/84534acbe04fe58f1874a03e1e1c5b43


Deledrius:
Excellent work!  I'm so happy to see this being tackled. :D

Hoikas:
I wonder if this red patch is here in TPotS?
image

The image is looking at the linking book from just outside of the "ring" - I flew over there when I could just barely see the red color from the link in point.

Mystler:
That red spot is on the upper level mesh that i didn't touch at all but it's an interesting question. I don't think i have a TPotS save around to check.

Hoikas:
Darn, I was hoping you would... But I suddenly remembered, you can just link to AhnySphere02 using the offlineki's `/link` command. So I did:

image

Hoikas (approved):

